PR #326, opened 2020-03-25
Precision

jalelmahjoubi:


EmmanuelDemey:
@jalelmahjoubi what is this new .history folder ?

jalelmahjoubi:
> @jalelmahjoubi what is this new .history folder ?

src/model/transformations/codes-list.js
```diff
@@ -36,8 +38,9 @@ export function getcodelistwithclarification(remote, variableclarification) {
   remote.forEach(codelist => {
     variableclarification.forEach( clarif => {
       if (clarif.codelistid === codelist.id){
-          codelist.Code[clarif.position] = {
-            ...codelist.Code[clarif.position], 
+      const index = codelist.Code.findIndex(code => code.Value == clarif.position)
```
EmmanuelDemey, line 41: wrong indentation

src/model/transformations/collected-variable.js
```diff
@@ -3,13 +3,24 @@ import {
...
+import { element } from 'prop-types';
 const { COLLECTED } = VARIABLES_TYPES;
 
 export function remoteToStore(
   remote = [],
   responsesByVariable,
   codesListsStore,
+  variableclarification,
 ) {
+  remote.map(variable => {
```
EmmanuelDemey, line 15: a .map function should return something. 
If it is not the case, you should prefer .forEach

src/widgets/codes-lists/components/codes-lists-input-code.jsx
```diff
@@ -82,7 +82,7 @@ class CodesListInputCode extends ComponentWithValidation {
         change(formName, `${path}precisionlabel`, code.precisionlabel);
       }
       else if(precisionShow) { 
-        change(formName, `${path}precisionlabel`, `${Question}${code.value}CL label`);
+        change(formName, `${path}precisionlabel`, `Préciser :`);
```
EmmanuelDemey, line 85: I think the "Préciser :" string should come from the dictionary.

PR #331, opened 2020-05-05
Duration Format Hours/Hundredths

ritamds:
Ticket 192 - Ajouter un format pour les Réponse simple/Durée en heures-centièmes

This adds a new format for Response Simple of type Duration that is Hours/Hundredths (HH:CH), and adds unit tests for this format.
Also fixes a small bug in PTnHnM format code.

I create new variables for hours "mihundhours" and "mahundhours" instead of reusing the existing ones because they have different validation rules (values must be between 0 and 50). 
I also add a 0 to the left when the value is between 0 and 9 (for hours and hundredths) to be in accordance with the desired output in the ticket.
Everything else is done in a similar way to the existing code for the other Durations.


src/model/transformations/collected-variable.js
```diff
@@ -85,6 +89,10 @@ export function remoteToStore(
           datatype.mayears = matches_maximum[0] == 0 ? '' : matches_maximum[0];
           datatype.mamonths = matches_maximum[1] == 0 ? '' : matches_maximum[1];
         }
+        if (format !== undefined && format === 'HH:CH') {
```
EmmanuelDemey, line 92: Why do you need to test undefined && HH:CH ? The second test is not enough ?

src/model/transformations/response-format-simple.js
```diff
@@ -43,6 +43,10 @@ export function remoteToState(remote) {
         datatype.miyears = matches_minimum[0] == 0 ? '' : matches_minimum[0];
         datatype.mimonths = matches_minimum[1] == 0 ? '' : matches_minimum[1];
       }
+      if (format !== undefined && format === 'HH:CH') {
```
EmmanuelDemey, line 46: Same remark as before

src/model/transformations/response-format-simple.js
```diff
@@ -55,6 +59,10 @@ export function remoteToState(remote) {
         datatype.mayears = matches_maximum[0] == 0 ? '' : matches_maximum[0];
         datatype.mamonths = matches_maximum[1] == 0 ? '' : matches_maximum[1];
       }
+      if (format !== undefined && format === 'HH:CH') {
```
EmmanuelDemey, line 62: Same remark as before

src/model/transformations/response-format-simple.js
```diff
@@ -110,13 +122,21 @@ export function stateToRemote(state, collectedVariables) {
...
       }
       if(mahours !== "" || maminutes !== ""){
        durationDataType.maximum = `PT${mahours || 0}H${maminutes || 0}M`;
       }
     }
+    if (dataType.format === 'HH:CH') {
+      if(mihundhours || mihundredths){
+       durationDataType.minimum = `${('0' + mihundhours).slice(-2) || 0}:${('0' + mihundredths).slice(-2) || 0}`;
+      }
+      if(mahundhours !== "" || mahundredths !== ""){
```
EmmanuelDemey, line 136: single quotes

src/model/transformations/response-format-table.js
```diff
@@ -91,6 +91,10 @@ function getMeasuresModel(responses, dimensions, offset) {
           responses[i].Datatype.Miyears = matches_minimum[0] == 0 ? '' :matches_minimum[0];
           responses[i].Datatype.Mimonths = matches_minimum[1] == 0 ? '' :matches_minimum[0];
         }
+        if (responses[i].Datatype.Format !== undefined && responses[i].Datatype.Format === 'HH:CH') {
```
EmmanuelDemey, line 94: Same remark as before

src/model/transformations/response-format-table.js
```diff
@@ -103,6 +107,10 @@ function getMeasuresModel(responses, dimensions, offset) {
           responses[i].Datatype.Mayears = matches_maximum[0] == 0 ? '' :matches_maximum[0];
           responses[i].Datatype.Mamonths = matches_maximum[1] == 0 ? '' :matches_maximum[0];
         }
+        if (responses[i].Datatype.Format !== undefined && responses[i].Datatype.Format === 'HH:CH') {
```
EmmanuelDemey, line 110: Same remark as before

src/widgets/component-new-edit/components/response-format/simple/simple-duree.jsx
```diff
@@ -58,9 +58,16 @@ class ResponseFormatDatatypeDuree extends Component {
...
+              key="HH:CH"
+              value="HH:CH"
+            >
+              {Dictionary.durationformat3}
+            </GenericOption>
+
           </Field>   
           <div className="response-format-datatype-duree-minimum" >
-            <div className="response-format-datatype-duree-label-minimum" style= {{ display : formatini==='PTnHnM' || formatini==='PnYnM'? 'flex' : 'none' }}>   
+            <div className="response-format-datatype-duree-label-minimum" style= {{ display : formatini==='PTnHnM' || formatini==='PnYnM' || formatini==='HH:CH' ? 'flex' : 'none' }}>
```
EmmanuelDemey, line 70: We can maybe extract this test in a dedicated method ? In order to make this HTML easier to read/ What do you think ?

src/widgets/component-new-edit/components/response-format/simple/simple-duree.jsx
```diff
@@ -101,11 +108,31 @@ class ResponseFormatDatatypeDuree extends Component {
...
+                    disabled={this.props.readOnly}
+                />
+            </div>
+
             </div>
           </div>
 
           <div className="response-format-datatype-duree-maximum" >
-            <div className="response-format-datatype-duree-label-maximum" style= {{ display : formatini==='PTnHnM' || formatini==='PnYnM' ? 'flex' : 'none' }}>   
+            <div className="response-format-datatype-duree-label-maximum" style= {{ display : formatini==='PTnHnM' || formatini==='PnYnM' || formatini==='HH:CH' ? 'flex' : 'none' }}>
```
EmmanuelDemey, line 135: Same remark as before

src/model/transformations/collected-variable.js
ritamds: The test was already part of existing code, but I believe the second test is enough, yes. With and without the test the result is the same, so I have removed it.

src/widgets/component-new-edit/components/response-format/simple/simple-duree.jsx
ritamds: I propose a solution of creating a variable `isDuration` and assigning it this test, allowing us to simplify the style to `display : isDuration ? 'flex' : 'none'`. What do you think?
I believe this way we can achieve better code readability and remain within the scope of this ticket. Ultimately, it could maybe be beneficial to have a solution to minimise the repetition of the style attribute (could be a method), perhaps having a specific ticket for code refactoring?

PR #339, opened 2020-05-18
Manage NoDataByDefinition

ritamds:
Ticket 312B - Gestion des NoDataByDefinition

This adds the checkbox `Collected` in the tab 'Collected Variables' in responses of type `Table`.
This checkbox is checked by default. When unchecked by the user, an object, with keys `AttributeValue` and `AttributeTarget` relating to that variable, is added to `Attribute` in the JSON.
Also fixes some typos in the existing code.

### DIFF
--- a/src/model/transformations/responses.js
+++ b/src/model/transformations/responses.js
@@ -8,8 +8,8 @@ export function stateToModel(
   type,
 ) {
   let collectedVariablesFinal = collectedVariables;
-  if(type == 'MULTIPLE_CHOICE' || type == 'TABLE') {
-    collectedVariablesFinal.map(collected=> {
+  if(type === QUESTION_TYPE_ENUM.MULTIPLE_CHOICE || type === QUESTION_TYPE_ENUM.TABLE) {
+    collectedVariablesFinal.map(collected => {
       const find = Object.values(collectedVariablesStore).find(variable=> variable.id == collected);
       if(find && find.type == 'TEXT' && !find.x) {
         collectedVariablesFinal = collectedVariablesFinal.filter(element => element != find.id)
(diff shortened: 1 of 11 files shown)


src/model/transformations/responses.js
```diff
@@ -19,17 +19,34 @@ export function stateToModel(
...
 
+  let attributeModel = [];
+
+  const mappingModel = responsesModel.map(r => {
+    const { x, y, isCollected } = collectedVariablesStore[r.CollectedVariableReference];
     // Table : Fix lines and look into columns
     const MappingTarget =
       type === QUESTION_TYPE_ENUM.MULTIPLE_CHOICE ? `${x}` : `${x} ${y}`;
+    
+      if (isCollected === false ) {
```
EmmanuelDemey, line 31: Why you do not use !isCollected ?

src/model/transformations/responses.js
```diff
@@ -19,17 +19,34 @@ export function stateToModel(
...
+          "AttributeTarget": MappingTarget
+        },)
+      }
     return { MappingSource: r.id, MappingTarget };
   });
 
-  return {
-    Response: responsesModel,
-    Mapping: mappingModel,
+  if (type == 'TABLE') {
```
EmmanuelDemey, line 40: I think we have a const somewhere for this string.

src/widgets/component-new-edit/components/variables/collected-variables.jsx
```diff
@@ -99,6 +99,8 @@ class CollectedVariables extends Component {
       componentName,
       responseFormatType,
     } = this.props;
+    const hiddenCollected = responseFormatType !== "TABLE";
```
EmmanuelDemey, line 102: You should use the const we already defined in the application

src/model/transformations/responses.js
ritamds: Changed.

src/model/transformations/responses.js
ritamds: Yes, we have. I have replaced with it now.

src/widgets/component-new-edit/components/variables/collected-variables.jsx
ritamds: Changed.

ritamds: Done!

PR #342, opened 2020-05-28
Add tooltips to Measure Label Input

ritamds:
306-FO-associer ou supprimer des info-bulles à des libellés de colonne de tableaux

This changes the input `Measure Label` in answers of type `Table` to `RichTextareaWithVariableAutoCompletion`, and uses the new `toolbarConfigTooltip` that displays only the tooltips.
The toolbar is passed as prop. If no toolbar prop is passed, it uses the default `toolbarConfig`.

### DIFF
--- a/src/widgets/component-new-edit/components/response-format/table/input-measure.jsx
+++ b/src/widgets/component-new-edit/components/response-format/table/input-measure.jsx
@@ -3,7 +3,7 @@ import { Field } from 'redux-form';
 import PropTypes from 'prop-types';
 
 import { RichTextareaWithVariableAutoCompletion } from 'forms/controls/control-with-suggestions';
-import {toolbarConfigTooltip} from "forms/controls/rich-textarea";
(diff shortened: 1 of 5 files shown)


src/widgets/component-new-edit/components/response-format/table/input-measure.jsx
```diff
@@ -2,7 +2,8 @@ import React from 'react';
 import { Field } from 'redux-form';
 import PropTypes from 'prop-types';
 
-import { InputWithVariableAutoCompletion } from 'forms/controls/control-with-suggestions';
+import { RichTextareaWithVariableAutoCompletion } from 'forms/controls/control-with-suggestions';
+import {toolbarConfigTooltip} from "forms/controls/rich-textarea";
```
EmmanuelDemey, line 6: You should use single quote.

ritamds: Done!
I didn't notice, thank you for noticing!

PR #343, opened 2020-06-01
Loading Message when Fetching Questionnaires

ritamds:
318FO-Message au chargement de la liste de questionnaires en page d'accueil

This renders a loading message when fetching questionnaires.

### DIFF
--- a/src/layout/questionnaire-list/components/questionnaire-list.jsx
+++ b/src/layout/questionnaire-list/components/questionnaire-list.jsx
@@ -34,17 +34,25 @@ class QuestionnaireList extends Component {
     super(props);
     this.state = {
       filter: '',
+      loaded: false,
     };
   }
 
   UNSAFE_componentWillMount() {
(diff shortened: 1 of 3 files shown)


src/layout/questionnaire-list/components/questionnaire-list.jsx
```diff
@@ -34,17 +34,25 @@ class QuestionnaireList extends Component {
     super(props);
     this.state = {
       filter: '',
+      loaded: false,
     };
   }
 
   UNSAFE_componentWillMount() {
-    if (this.props.user && this.props.user.permission)
+    if (this.props.user && this.props.user.permission){
```
EmmanuelDemey, line 42: As you are working on this method, this is maybe the best time to remove the UNSAFE-prefixed method. You should have a look to the react documentation in order to know what is the new syntax.

ritamds: Agreed. I will work on it.

PR #344, opened 2020-06-01
Loops

jalelmahjoubi:


src/actions/component-remove.js
```diff
@@ -25,9 +26,9 @@ export function removeComponentFromActivesComponent(
     if (currentId !== deletedComponent.id) {
       acc[currentId] = {
         ...activesComponents[currentId],
-        children: activesComponents[currentId].children.filter(
+        children: activesComponents[currentId] && activesComponents[currentId].children? activesComponents[currentId].children.filter(
```
EmmanuelDemey, line 29: Missing space. 
 
In a new version of Babel, we can now use the elvis operator ?. . 

This operator is usefull if we want to avoid this kind of syntax. 

We should maybe update babel dependencies if doable.

src/actions/component-remove.js
```diff
@@ -256,8 +257,11 @@ export function removeSequence(activesComponents, deletedComponent) {
  */
 export function remove(activesComponents, idDeletedComponent) {
   const deletedComponent = activesComponents[idDeletedComponent];
-
-  if (deletedComponent.children.length === 0) {
+  
+  if(isLoop(deletedComponent)) {
+    return removeComponentFromActivesComponent(activesComponents, deletedComponent);
+  }
+  if (!isLoop(deletedComponent) && deletedComponent.children.length === 0) {
```
EmmanuelDemey, line 264: I think `!isLoop(deletedComponent)` is useless because you have the opposite if right before with a return statement,

src/layout/generic-input/components/generic-input.jsx
```diff
@@ -102,20 +120,30 @@ class GenericInput extends Component {
...
         ...this.state,
-        showNewUnsavedModal: true,
+        showNewLoopModal: true,
       };
       this.setState(newState);
     }
-   })
+    else {
+      this.props.saveActiveQuestionnaire().then(()=>{
+        if(this.props.isQuestionnaireHaveerror){
```
EmmanuelDemey, line 138: `isQuestionnaireHaveerror` should be replaced with `isQuestionnaireHaveError`

src/layout/generic-input/utils/generic-input-utils.js
```diff
@@ -261,3 +260,21 @@ export function getNewQuestionPlaceholder(components, activeComponent) {
...
+/**
+ * Get new loop placeholder
+ *
+ * It finds from a list of components if it have at least one sequence
+ *
+ * @param  {object} components        List of components
+ * @return {string} true if component have a sequence
+ */
+export function getNewLoopPlaceholder(components) {
+  let componentHaveSequence = true;
```
EmmanuelDemey, line 273: I think all this method can be replaced with something like : 
`return Object.values(components).find(component => component.type === SEQUENCE);`

src/model/transformations/collected-variable.js
```diff
@@ -149,9 +236,28 @@ export function storeToRemote(store) {
...
       },
     };
-    if(!z) {
+    
+    const questionsInLoop =  findQuestionInLoop(componentsStore);
+    const collectedScop = getCollectedScope(questionsInLoop, id);
+
+    if(collectedScop.component) {
+       if(collectedScop.component.type === QUESTION && 
+          collectedScop.component.responseFormat.type === "TABLE"
```
EmmanuelDemey, line 245: We do not have constant like QUESTION for "TABLE" and "LIST" ?

src/widgets/component-new-edit/components/variables/calculated-variables.jsx
```diff
@@ -39,9 +42,17 @@ export const defaultProps = {
   errors: [],
 };
 
-// Component
-
-function CalculatedVariables({ formName, selectorPath, errors, addErrors }) {
+function CalculatedVariables({ formName, selectorPath, errors, addErrors, componentsStore }) {
+  const scopeOption = Object.values(componentsStore)
+  .filter(component=> component.type === "LOOP" && !component.basedOn || component.type === "LOOP" && component.basedOn && componentsStore[component.basedOn] && componentsStore[component.basedOn].type != "LOOP" )
```
EmmanuelDemey, line 47: you should use a const for the string "LOOP"
This line is way too long. you should first wrap your tests between ( ) because for exemple you have twice this test : `component.type === "LOOP"`.

src/widgets/component-new-edit/components/variables/calculated-variables.jsx
```diff
@@ -39,9 +42,17 @@ export const defaultProps = {
...
+  .map(element => {
+  return (<GenericOption
```
EmmanuelDemey, line 49: wrong indentation

src/widgets/component-new-edit/components/variables/external-variables.jsx
```diff
@@ -41,7 +44,17 @@ export const defaultProps = {
 
 // Component
 
-function ExternalVariables({ formName, selectorPath, errors, addErrors }) {
+function ExternalVariables({ formName, selectorPath, errors, addErrors, componentsStore }) {
+  const scopeOption = Object.values(componentsStore)
+  .filter(component=> component.type === "LOOP" && !component.basedOn || component.type === "LOOP" && component.basedOn && componentsStore[component.basedOn] && componentsStore[component.basedOn].type != "LOOP" )
```
EmmanuelDemey, line 49: Same comment as before 👆
If this is exactly the same code, we can maybe extract into a method, that you'll be able to unit test.

src/widgets/component-new-edit/components/variables/external-variables.jsx
```diff
@@ -41,7 +44,17 @@ export const defaultProps = {
...
+  .map(element => {
+  return (<GenericOption
```
EmmanuelDemey, line 51: Wrong indentation

src/layout/generic-input/utils/generic-input-utils.spec.js
```diff
@@ -195,7 +195,7 @@ describe('getNewLoopPlaceholder', () => {
 
     };
     const result = getNewLoopPlaceholder(components);
-    expect(result).toEqual(true);
+    expect(result).toEqual(undefined);
```
EmmanuelDemey, line 198: If you want to convert to a boolean value, you can two extra !! to the implementation. 

return !!Object.values(components).find(component => component.type === SEQUENCE);

PR #349, opened 2020-06-08
Redirection end

jalelmahjoubi:


src/model/transformations/component.js
```diff
@@ -259,21 +258,23 @@ function remoteToStoreNested(
...
-      { ...child, weight, parent },
-      componentGroup,
-      codesListsStore,
-    );
-    weight += 1;
-    if (child.Child)
-      remoteToStoreNested(
-        child.Child,
-        child.id,
+    if(child.Name !== "QUESTIONNAIRE_END") {
```
EmmanuelDemey, line 261: Could we use a constant instead of this hard-coded string ?

src/model/transformations/questionnaire.js
```diff
@@ -145,6 +148,21 @@ export function stateToRemote(state, stores) {
...
+    Name: "QUESTIONNAIRE_END",
+    Label: [
+      "QUESTIONNAIRE_END"
+    ],
+    Declaration: [],
+    genericName: "MODULE",
+    Control: [],
+    FlowControl: [],
+    TargetMode: TargetMode,
+    type: "SequenceType",
```
EmmanuelDemey, line 163: I think we alread have constants for this hard-coded strings

PR #362, opened 2020-06-24
Remove Parsing Variables with Single Dollar

ritamds:
263 -FO- Retirer la parsing sur les variables avec un seul dollar (générateur DDI)

This checks for variables in labels, upon Validation, and, if they have a single $ (e.g. `$VAR`), adds a $ at the end (`$VAR$`).

### DIFF
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -70,10 +70,16 @@ export function updateNameField(currentValueLabel, currentValueName) {
 }
 
 export function verifyVariable(label) {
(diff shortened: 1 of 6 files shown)


src/utils/utils.js
```diff
@@ -68,3 +68,19 @@ export function updateNameField(currentValueLabel, currentValueName) {
 
   return value;
 }
+
+export function verifyVariable(label) {
```
EmmanuelDemey, line 72: You should write an unit test for this new function.

PR #366, opened 2020-06-29
[FIX] Loading Message 

ritamds:
318-FO- Message au chargement de la liste de questionnaires en page d'accueil

This moves the logic of the `Loading Message` from the `State` to the `Redux Store`.

### DIFF
--- a/src/layout/questionnaire-list/components/questionnaire-list.jsx
+++ b/src/layout/questionnaire-list/components/questionnaire-list.jsx
@@ -34,25 +34,18 @@ class QuestionnaireList extends Component {
     super(props);
     this.state = {
       filter: '',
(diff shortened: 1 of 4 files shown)


EmmanuelDemey:
You should revert this change. You should keep the first implementation, but your should set the loaded state after the promises is resolved. Something like this snippet :  

```
this.props.loadQuestionnaireList(nextProps.user.permission);	      this.props.loadQuestionnaireList(nextProps.user.permission)
.then(() => {
   this.setState({	
        loaded: true,	
    });
});
```

ritamds:
@EmmanuelDemey Thank you. I will close this PR and do a new one from the previous code.

PR #384, opened 2020-10-01
Drag drob filters

jalelmahjoubi:


src/actions/component.js
```diff
@@ -232,6 +234,11 @@ export const updateComponent = (
     },
   };
 };
+
+// const modifieFilters = idTargetComponent => (dispatch, getState) => {
```
EmmanuelDemey, line 238: You should not put into GIT commented code. You can remove it

src/actions/component.js
```diff
@@ -246,14 +253,100 @@ export const dragComponent = (
 ) => (dispatch, getState) => {
   const state = getState();
   const activesComponents = state.appState.activeComponentsById;
+  const find_Filters = Object.values(activesComponents).filter(
+    element =>
+      element.type === FILTER &&
+      (element.initialMember === idTargetComponent ||
+        element.finalMember === idTargetComponent),
+  );
+  if (find_Filters?.length > 0) {
```
EmmanuelDemey, line 262: I think find_Filters won't never be undefined. So you can remove the ?

src/actions/component.js
```diff
@@ -246,14 +253,100 @@ export const dragComponent = (
...
+    find_Filters.forEach(filter => {
+      if (
+        filter.initialMember === idTargetComponent &&
```
EmmanuelDemey, line 265: I think the statement `filter.initialMember === idTargetComponent` is unnecessary.

EmmanuelDemey (approved):

PR #406, opened 2021-01-04
Refactoring

jalelmahjoubi:


build-config/fake-backend/server.js
```diff
@@ -53,6 +53,17 @@ server.use(restifyQueryParser());
 
 server.get('/questionnaires/search', function(req, res, next) {
   // @TODO: Take into account the property "owner"
+  const questionnaires1 = [];
+  questionnaires.forEach(question => {
```
EmmanuelDemey, line 57: You can maybe use a .map instead of a .forEach/.push

EmmanuelDemey (changes requested):

src/layout/page-questionnaire/components/page-questionnaire.jsx
```diff
@@ -69,6 +72,12 @@ const PageQuestionnaire = props => {
       props.loadQuestionnaire(id);
       setIdState(id);
     }
+    console.log('loading', loading);
+
+    // if (loading !== loadingState) {
```
EmmanuelDemey, line 77: If this code is not used anymore, you should remove it

src/layout/page-questionnaire/components/page-questionnaire.jsx
```diff
@@ -135,9 +145,21 @@ const PageQuestionnaire = props => {
 
   return (
     <div id={COMPONENT_ID}>
-      <QuestionnaireNav />
-      <QuestionnaireListComponents navigate={props.history.push} />
-      <GenericInput />
+      {loading !== false ? (
```
EmmanuelDemey, line 148: Why not `!loading` ?

src/scss/_global.scss
```diff
@@ -308,6 +308,12 @@ a {
   display: flex;
 }
 
+.loaderClass {
+  position: fixed;
+top: 50%;
```
EmmanuelDemey, line 313: Wrong indentation

EmmanuelDemey (changes requested):

EmmanuelDemey (approved):

build-config/fake-backend/server.js
```diff
@@ -54,7 +54,7 @@ server.use(restifyQueryParser());
 server.get('/questionnaires/search', function(req, res, next) {
   // @TODO: Take into account the property "owner"
   const questionnaires1 = [];
-  questionnaires.forEach(question => {
+  questionnaires.map(question => {
```
EmmanuelDemey, line 57: That's not I was looking for. Something like that : 
```
const questionnaires1 = questionnaires.map(question => {
   return {
      label: question.Label,	      label: question.Label,
      ...
  }
})
```

EmmanuelDemey (changes requested):

PR #428, opened 2021-02-15
'modification-popup-setup'

jalelmahjoubi:


src/layout/generic-input/components/generic-input.jsx
```diff
@@ -160,6 +160,7 @@ class GenericInput extends Component {
       componentIdForPageBreak,
       activeQuestionnaire,
     } = this.props;
+    console.log('isQuestionnaireModified', isQuestionnaireModified)
```
loichenninger, line 163: I think you can remove this console.log

loichenninger (approved):

PR #449, opened 2021-03-18
'fusion-2Same-Question'

jalelmahjoubi:


src/layout/questionnaire-list/components/questionnaire.jsx
```diff
@@ -37,28 +37,30 @@ function QuestionnaireListItem({
...
-          onClick={event => {
-            event.preventDefault();
-            handleOpenPopup(id, label);
-          }}
-        >
-          {Dictionary.duplicate}
-        </button>
+        <div className="check-button">
+          {fusion ? (
+            <input
```
EmmanuelDemey, line 42: You should add an aria-label to this input in order to define a textual alternative.

EmmanuelDemey (changes requested):

EmmanuelDemey (approved):

PR #450, opened 2021-03-19
Delete id filtre

jalelmahjoubi:


src/model/transformations/redirection-filters.js
```diff
@@ -17,7 +17,7 @@ function getFiltersState(filter, parent, type) {
   );
   return {
     id,
-    name,
+    name: name || false,
```
EmmanuelDemey, line 20: Why not a empty string instead of setting a boolean value to a string variable ?

EmmanuelDemey (changes requested):

EmmanuelDemey (approved):

EmmanuelDemey (approved):

PR #576, opened 2022-03-31
Adding spinner when visualizing

loichenninger:


BulotF (approved):
dictionary : Occurred has 2 R

PR #731, opened 2024-02-13
Dev check variable evolution in array

BulotF:
Refactor Question validation rules + collected variable initialisation


laurentC35 (approved):
These function seams to be cognitively complex.

We need to refactor this one day

Issue #695, opened 2023-09-13
Dynamic Unit on InputNumber


laurentC35:
## ToDo

![image](https://github.com/InseeFr/Pogues/assets/38245508/b8695398-5257-4993-84cf-0018d1e3f847)

Eno : https://github.com/InseeFr/Eno/issues/715
Lunatic: https://github.com/InseeFr/Lunatic/issues/668

romaintailhurat:
As mentioned in previous issues, i would prefer to have in this case an exclusive path, like a switch button for choosing between static and dynamic measure units?

But most importantly, if we allow for dynamic, what is the use for a static unit list? Most designers would just use the dynamic option and write whatever they like as units. 🤔 

So probably having dynamic measure units means deleting static ones, maybe at the expense of harmonization.

QRuhier:
⚠️ impacts on Pogues-model & ddi


Currently in jsonPogues, unit is actually an url, used in transformation to ddi for getting the value of the unit.
```
{
        "id": "lypl2fz1",
        "Name": "QUESTIONNO",
        "type": "CollectedVariableType",
        "Label": "QUESTIONNO label",
        "Datatype": {
          "Unit": "url_of_unit",
          "type": "NumericDatatypeType",
          "Maximum": "10",
          "Minimum": "1",
          "Decimals": "",
          "typeName": "NUMERIC"
        }
      }
```
If we want to make the difference between VTL expression or url (that leads to a different transformation), we need to change Pogues-Model

Then, in ddi, we do have 
```
<l:VariableRepresentation>
               <r:NumericRepresentation>
                  <r:MeasurementUnit>€</r:MeasurementUnit>
                  <r:NumberRange>
                     <r:Low isInclusive="true">1</r:Low>
                     <r:High isInclusive="true">10</r:High>
                  </r:NumberRange>
                  <r:NumericTypeCode controlledVocabularyID="INSEE-CIS-NTC-CV">Decimal</r:NumericTypeCode>
               </r:NumericRepresentation>
            </l:VariableRepresentation>
````
We cannot directly give the expression in <r:MeasurementUnit>

**Comment by laurentC35** [comment]
Modélisation:

```
<r:MeasurementUnit>¤VAR_COLLECTED¤</r:MeasurementUnit>
```
ou

```
<r:MeasurementUnit>€</r:MeasurementUnit>
```

QRuhier:
Because of DDI modelisation, for dynamic unit we must save as value the id of a variable.
[...]

PR #803, opened 2024-09-11
feat: dynamic unit on input number

QRuhier:
- #695


src/constants/dictionary.jsx
```diff
@@ -521,10 +521,18 @@ const dictionary = {
     en: 'Please respect the date format',
     fr: 'Merci de respecter le format de la date',
   },
+  dynamicUnit: {
+    en: 'Custom unit of measure',
```
chloe-renaud, line 525: Customized unit of measurement

src/constants/dictionary.jsx
```diff
@@ -521,10 +521,18 @@ const dictionary = {
...
+    fr: 'Unité de mesure personnalisée',
+  },
   unit: {
     en: 'Unit of measure',
     fr: 'Unité de mesure',
   },
+  dynamicUnitFormula: {
+    en: 'Formula of the unit of measure',
```
chloe-renaud, line 533: Unit of measurement formula

src/model/transformations/response.spec.jsx
```diff
@@ -135,7 +135,7 @@ describe('response tranformations', () => {
     expect(result.Datatype.Maximum).toEqual(maximum);
   });
   test('when Decimals is defined', () => {
-    const typeName = 'DATE';
```
chloe-renaud, line 138: Why is it not `'DATE'` anymore?

src/widgets/component-new-edit/components/response-format/simple/simple-numeric.jsx
```diff
@@ -44,14 +58,42 @@ function ResponseFormatDatatypeNumeric({ name, required, readOnly }) {
...
+          component={ListRadios}
+          label={Dictionary.dynamicUnit}
           disabled={readOnly}
-        />
+          onChange={handleDynamicUnitChange}
+          required
+          // Convert string "true"/"false" to boolean true/false when storing in Redux form
+          parse={value => value === 'true'}
+          // Convert boolean true/false to string "true"/"false" when displaying the form
+          format={value => (value === true ? 'true' : 'false')}
```
chloe-renaud, line 71: To simplify we could have defined this function as:
```
  value => `${value}`
```
But to each their own

src/widgets/component-new-edit/components/response-format/simple/simple-numeric.jsx
```diff
@@ -44,14 +58,42 @@ function ResponseFormatDatatypeNumeric({ name, required, readOnly }) {
...
+        >
+          <GenericOption key="1" value="true">
```
chloe-renaud, line 73: If the `key` is meant as an React `key` to identify an item in a list, it is better practice to use some sort of `id` instead of an index (eg `"dictionary-true"` instead of `"1"`) since the `key` id should never change for each item.

See: https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key

Here it doesn't seem to be useful since it's not an array though so we should remove this property since `key` does not seem to be used by the `GenericOption` component.

src/model/transformations/response.spec.jsx
QRuhier: it actually does not change anything because we manage it in the same way whatever the type

src/widgets/component-new-edit/components/response-format/simple/simple-numeric.jsx
QRuhier: i'll change the comment, actually value can be undefined and i want to handle it as a false value

chloe-renaud (approved):

Issue #741, opened 2024-03-13
Delete duration option "heures/centièmes"


romaintailhurat:
This option could be replaced by a number input with 2 decimals.

See https://github.com/InseeFr/Lunatic/issues/258

BulotF:
Le besoin a disparu de ACEMOTRIM entre 2020 et 2021 ; J'ai regardé que c'était absent des autres enquêtes ACEMO de 2024... On peut effectivement supprimer

QRuhier:
Ok, so we assume that we completely remove this duration type, and that any existing questionnaire using it could be unreadable

QRuhier:
When reading an existing questionnaire with hours/hundredths does not crash. Just when opening the question, the duration type is to be set again.

For checking an existing questionnaire with hours/hundredth, you can import this one : 
[issue_741.json](https://github.com/user-attachments/files/17221308/issue_741.json)

PR #861, opened 2024-10-02
feat: delete duration hours hundredths

QRuhier:
- #741 

We assume that we do not handle hours/hundredths anymore, even for reading in current model


QRuhier:
Here, reading an existing questionnaire with hours/hundredths does not crash. Just when opening the question, the duration type is to be set again

BulotF (approved):

BulotF (approved):

PR #884, opened 2024-12-09
Fix/cant create survey in demo

MailineN:
- Make series-related fields optional if no series are retrieved from the backend or mock data.
- Minor UI adjustment to indicate mandatory fields.

### DIFF
--- a/src/widgets/statistical-context-criteria/containers/statistical-context-criteria.jsx
+++ b/src/widgets/statistical-context-criteria/containers/statistical-context-criteria.jsx
@@ -86,7 +86,7 @@ export const mapStateToProps = (
             ...storeToArray(state?.metadataByType?.series),
           ]
         : storeToArray(state?.metadataByType?.series),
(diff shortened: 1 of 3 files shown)


src/widgets/statistical-context-criteria/containers/statistical-context-criteria.jsx
```diff
@@ -86,6 +86,7 @@ export const mapStateToProps = (
             ...storeToArray(state?.metadataByType?.series),
           ]
         : storeToArray(state?.metadataByType?.series),
+    required: state?.metadataByType?.series?.length > 0,
```
laurentC35, line 89: `state?.metadataByType?.series` field is a object (key, value).
So this object.length is always undefined.

A simply fix can be
```js
required: Object.keys(state?.metadataByType ?? {}).length > 0,
```

laurentC35 (changes requested):
A minor fix to do 😉

QRuhier (approved):

PR #897, opened 2025-01-08
feat: improved versions screen

chloe-renaud:
- Added description in versions screen
- Confirmation text is bigger
- Load button is yellow
- Changed date/name displayed

### DIFF
--- a/src/constants/dictionary.ts
+++ b/src/constants/dictionary.ts
@@ -1697,6 +1697,18 @@ const dictionary: Dictionary = {
     fr: 'Ouvrir',
     en: 'Open',
   },
+  versionsInfo: {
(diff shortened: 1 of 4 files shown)


src/constants/dictionary.ts
```diff
@@ -1697,6 +1697,18 @@ const dictionary: Dictionary = {
     fr: 'Ouvrir',
     en: 'Open',
   },
+  versionsInfo: {
+    fr: 'Pour afficher une précédente version, choisissez "Charger" puis valider. Cette version peut ensuite être modifiée et sauvegardée comme une nouvelle version.',
```
QRuhier, line 1701: maybe change 'Pour afficher' into 'Pour charger' ? (you say 'To load' in en trad)

QRuhier (approved):

PR #898, opened 2025-01-09
feat: remove unit on numeric calculated variable

QRuhier:
For calculated variables, if the response type is numeric : remove all the fields relative to its measurement unit.

---

As the model of numeric variable is shared for every numeric variable type (collected, external, calculated) and isDynamicUnit is mandatory, i voluntarily did not change the transformations (form/state/model).
So even if we cannot set a unit anymore, in the model we keep unit='' with isDynamicUnit=false (that were already the default values for every numeric variable type)

### DIFF
--- a/src/widgets/component-new-edit/components/response-format/simple/simple-numeric.jsx
+++ b/src/widgets/component-new-edit/components/response-format/simple/simple-numeric.jsx
@@ -1,3 +1,4 @@
+import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { Field, FormSection, change, formValueSelector } from 'redux-form';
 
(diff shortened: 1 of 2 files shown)


MailineN (approved):

src/widgets/component-new-edit/components/response-format/simple/simple-numeric.jsx
```diff
@@ -25,6 +25,7 @@ function ResponseFormatDatatypeNumeric({
   readOnly,
   isDynamicUnit,
   setUnit,
+  enableSetUnit,
```
chloe-renaud, line 28: Maybe add an explanation on why it is needed (business-rule).

chloe-renaud (changes requested):

chloe-renaud (approved):

PR #919, opened 2025-01-29
(PoC) new questionnaire page

chloe-renaud:
PoC tanstack query + tanstack form + toaster in create form page


chloe-renaud:
There is still a bit of work to do (i.e. display the error that prevent the form to be submitted) but I think we can do it later and merge as is for now.

QRuhier (approved):

PR #925, opened 2025-02-05
feat: display not found questionnaire

laurentC35:
fix tuleap#10956

### DIFF
--- a/src/api/questionnaires.ts
+++ b/src/api/questionnaires.ts
@@ -1,3 +1,5 @@
+import { HttpResponseError } from '@/errors/error';
+
 import { computeAuthorizationHeader, getBaseURI } from './utils';
 
 const pathQuestionnaire = 'persistence/questionnaire';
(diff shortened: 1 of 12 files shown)


src/api/questionnaires.ts
```diff
@@ -83,7 +83,10 @@ export async function getQuestionnaire(
 
   return fetch(url, {
     headers,
-  }).then((res) => res.json());
+  }).then((res) => {
+    if (res.ok) return res.json();
+    else throw new Error(`${res.status}`);
```
chloe-renaud, line 88: I don't like that our error has for message the status code. We could use either make a custom error (e.g. `NotFoundError` and we would check the e.name later on) or use the `cause` option to hand out the status code (e.g. `new Error("Questionnaire not found", { cause: res )`)

src/layout/page-questionnaire/components/page-questionnaire.jsx
```diff
@@ -113,11 +115,15 @@ const PageQuestionnaire = (props) => {
 
   return (
     <div id={COMPONENT_ID}>
-      {loading ? (
-        <Loader />
-      ) : (
+      {errorLoading && (
```
chloe-renaud, line 118: Would make more sense as `loadingError`. It's not obvious if it's supposed to be a boolean or an error otherwise.

chloe-renaud (approved):

PR #934, opened 2025-02-12
feat: add fr&en translations on next

MailineN:


QRuhier:
Could you look for improving type safety ?
- avoid to use a key that is not defined in dictionary
- have auto-completion with defined keys when using useTranslate()

chloe-renaud:
Can we handle form errors and toast translations?

https://github.com/InseeFr/Pogues/blob/6cefb071de9fbf3e47be283f9d139fc38b1f3399/next/src/components/createQuestionnaire/CreateQuestionnaire.tsx#L34

https://github.com/InseeFr/Pogues/blob/6cefb071de9fbf3e47be283f9d139fc38b1f3399/next/src/components/createQuestionnaire/CreateQuestionnaire.tsx#L111

next/src/i18n/locales/fr.ts
```diff
@@ -50,7 +50,7 @@ export const translations: Translations<'fr'> = {
     title: 'Questionnaires',
     create: 'Créer un questionnaire',
     lastUpdate: 'Dernière modification',
-    search: 'Rechercher un questionnaire TEST ETS ',
+    search: 'Rechercher un questionnaires',
```
chloe-renaud, line 53: Should be singular.

next/src/i18n/types.ts
```diff
@@ -0,0 +1,117 @@
...
+  commonMessage,
+  createQuestionnaireMessage,
+  externalMessage,
+  homeMessage,
+  questionnairesMessage,
+} from './componentKeys';
+import type { Language, fallbackLanguage } from './i18n';
+
+export type CommonTranslation = {
+  version: string;
```
chloe-renaud, line 15: Shouldn't we sort them alphabetically so that manual search and git history are handled better?

next/src/components/createQuestionnaire/CreateQuestionnaire.tsx
```diff
@@ -21,8 +21,10 @@ import {
 } from '@/models/questionnaires';
 import { uid } from '@/utils/utils';
 
+import { getTranslation, useTranslation } from '../../i18n';
```
chloe-renaud, line 24: Should be imported with `@` alias to avoid `../..` since we target a root import (i.e. `@/i18n`)

next/src/i18n/i18n.ts
```diff
@@ -1,24 +1,29 @@
-import i18n from 'i18next';
-import LanguageDetector from 'i18next-browser-languagedetector';
-import { initReactI18next } from 'react-i18next';
+import { createI18nApi, declareComponentKeys } from 'i18nifty';
```
chloe-renaud, line 1: It's not `react-i18next` anymore? :(

chloe-renaud (changes requested):

next/src/components/createQuestionnaire/CreateQuestionnaire.tsx
```diff
@@ -31,10 +33,8 @@ interface FormValues {
 }
 
 const questionnaireSchema = z.object({
-  title: z.string().min(1, 'You must provide a title'),
-  targetModes: z
-    .set(z.nativeEnum(TargetModes))
-    .min(1, 'You must select at least one target mode'),
+  title: z.string().min(1, t('titleMessage')),
```
chloe-renaud, line 36: Does it update when we change the language?

next/src/i18n/@types/resources.ts
```diff
@@ -0,0 +1,7 @@
+import en from '../locales/en.json';
+
+const resources = {
+  en,
+} as const;
```
QRuhier, line 5: For type-safety working, should be
`const resources = {
  translation: en,
} as const;`

QRuhier (changes requested):

QRuhier (approved):

PR #940, opened 2025-02-14
refactor: axios and more state of the art tanstack query in route loader

chloe-renaud:


next/src/api/questionnaires.ts
```diff
@@ -3,148 +3,109 @@ import { queryOptions } from '@tanstack/react-query';
...
-  headers.append('Authorization', computeAuthorizationHeader(token));
-
-  return fetch(url, {
-    method: 'PUT',
-    headers,
-    body: JSON.stringify(qr),
-  }).then((res) => {
-    if (res.ok) return res;
-    throw new Error(`Network request failed :${res.statusText}`);
+  return instance.put(`/peristence/questionnaire/${id}`, qr, {
```
QRuhier, line 84: typo in path `peristence` -> `persistence`

QRuhier (changes requested):

next/src/api/questionnaires.ts
chloe-renaud: Good catch

PR #961, opened 2025-03-11
chore: remove pattern field from front-end & model

MailineN:


next/src/api/models/pogues.ts
```diff
@@ -486,7 +486,6 @@ type SuggesterQueryParser = {
 type SuggesterQueryParserParams = {
   language?: string;
   min?: number;
-  pattern?: string;
```
QRuhier, line 489: should not be removed in this type. Not the same object

laurentC35 (approved):

PR #967, opened 2025-03-21
feat: remove VTL editor in code card label field & reorder declarations fields

QRuhier:
For declarations : 
- reorder fields (set the type first)
- for code cart, change the label field from VTL editor into a simple input

The goal is to simplify removing VTL, and to avoid that the user uses quotes in the label, because they are not handled well then by Eno

---
Some difficulties while adding tests :

- I did not find a way for properly retrieving form state values, if someone has a better idea than my weird mock.

- Without mocking the component using the vtl editor, i had this error : 
`TypeError: Unknown file extension ".css" for C:\Users\PWF0UQ\Documents\git projects\Bowie\Pogues\node_modules\@making-sense\antlr-editor\node_modules\monaco-editor\esm\vs\editor\standalone\browser\standalone-tokens.css `
I did not find a better way, so for now i decided to mock it as an input rendering only what i needed

### DIFF
--- a/src/widgets/component-new-edit/components/declarations.jsx
+++ b/src/widgets/component-new-edit/components/declarations.jsx
@@ -12,6 +12,7 @@ import {
 } from '../../../constants/pogues-constants';
 import { RichEditorWithVariable } from '../../../forms/controls/control-with-suggestions';
 import GenericOption from '../../../forms/controls/generic-option';
+import Input from '../../../forms/controls/input';
 import ListCheckboxes from '../../../forms/controls/list-checkboxes';
 import Select from '../../../forms/controls/select';
 import {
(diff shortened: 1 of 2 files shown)


QRuhier:
In my tests I did not find a way for properly retrieving form state values, if someone has a better idea than my weird mock

QRuhier:
Without mocking the component using the vtl editor, i had this error : 
`TypeError: Unknown file extension ".css" for C:\Users\PWF0UQ\Documents\git projects\Bowie\Pogues\node_modules\@making-sense\antlr-editor\node_modules\monaco-editor\esm\vs\editor\standalone\browser\standalone-tokens.css `

I did not find a better way, so for now i decided to mock it as an input rendering only what i needed

MailineN (approved):

PR #974, opened 2025-03-25
feat: display a warning to remind min and max rules on loop create

chloe-renaud:


QRuhier (approved):

src/constants/dictionary.ts
```diff
@@ -882,6 +882,10 @@ const dictionary: Dictionary = {
     fr: 'Pour plus de précision, se référer à la documentation.',
     en: 'For more information check the documentation.',
   },
+  loopMinMaxHouseholdContextWarning: {
+    fr: "Rappel : pour les enquêtes en contexte entreprises, le nombre d'occurrences minimum et maximum d'une boucle doivent être égaux.",
+    en: 'Reminder: for business surveys, the min and max number of occurences of a loop must be equal.',
```
MailineN, line 887: "... occurrences in a loop must be the same." sounds better 
I don't know if it's significant or not, but you might want to use single quotes instead of double quotes for the fr label

MailineN (approved):

src/constants/dictionary.ts
chloe-renaud: >  I don't know if it's significant or not, but you might want to use single quotes instead of double quotes for the fr label

The double quote are put on by default by prettier since there is single quote in the sentence (because of `d'une` et `d'occurences`)

PR #980, opened 2025-03-28
fix: set warn as the default criticity in controls field

MailineN:


QRuhier:
Could also be interesting to check here : https://github.com/InseeFr/Pogues/blob/next/src/model/formToState/component-new-edit/__mocks__/control.jsx

First const is not even used so can be removed.
The other ones are used only for tests. We could improve them with the 3 criticities for improving tests

QRuhier (approved):

PR #981, opened 2025-03-28
feat: add search code list by label or id

MailineN:
I also added a component that lists the filters currently in use to generate the result. It might be overkill for now but could become more useful as more filters are added.

I'm not entirely sure about the UI/UX, so I'm open to any suggestions.

Screenshot 2025-03-28 162508


next/src/components/questionnaires/overview/Questionnaires.tsx
```diff
@@ -26,17 +28,57 @@ export default function Questionnaires({
   questionnaires = [],
 }: Readonly<QuestionnairesProps>) {
   const { t } = useTranslation();
-  const [filter, setFilter] = useState<string>('');
+  const searchParams = new URLSearchParams(window.location.search);
```
chloe-renaud: We already have `selectedStamp` props which get this information directly from the router.

next/src/models/filter.ts
```diff
@@ -0,0 +1,11 @@
+// To be completed once we have more filters
+export enum FilterEnum {
```
chloe-renaud: I don't like naming an enum "enum". It's already known since it's its type. It would make more sense to name it `FilterTypes` instead imo.

next/src/models/filter.ts
```diff
@@ -0,0 +1,11 @@
...
+  Search = 'search',
+  Stamp = 'stamp',
+}
+
+export type Filter = {
+  filterType: FilterEnum;
```
chloe-renaud: It's kind of obvious it's a "filter" type since we're in the Filter type. We can simply name it `type`. Ditto below.

next/src/models/filter.ts
```diff
@@ -0,0 +1,11 @@
...
+  filterContent: string;
+  clearFilterFunction?: () => void;
```
chloe-renaud: It's obvious it's a function from its type, and it should be from its naming (a verb). I would simply name it `clear`.

next/src/components/ui/icons/CancelIcon.tsx
chloe-renaud, line 1: Could be good practice to add a comment to explain what an icon looks like and when it should be use.

e.g.
```
/** Icon of an 'X' which should be used when one wants to close something or remove an item in a badge. */
export default function CloseIcon({}) { ... }
```

next/src/components/ui/Input.tsx
```diff
@@ -22,10 +28,15 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
       required = false,
       style = {},
       defaultValue,
+      onClear,
+      showClearButton = false,
       ...props
-    }: Readonly<InputProps>,
+    }: Readonly<
+      InputProps & { onClear?: () => void; showClearButton?: boolean }
```
chloe-renaud: There shouldn't be a need to re-add your props here since they're defined lines 16-17.

next/src/components/ui/Input.tsx
```diff
@@ -40,12 +51,23 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
...
-          {...props}
-        />
+        <div className="relative w-full">
+          <UIInput
+            ref={ref}
+            required={required}
+            className="w-full text-sm font-sans font-normal p-4 rounded-lg shadow-xs border border-default hover:enabled:border-primary focus:enabled:border-primary bg-default text-default placeholder:text-placeholder disabled:text-disabled disabled:bg-disabled focus-visible:outline focus-visible:outline-1 focus-visible:outline-primary"
+            {...props}
+          />
+          {showClearButton && (
```
chloe-renaud, line 62: Maybe we could simply check if "onClear" is defined.

next/src/components/ui/Input.tsx
```diff
@@ -2,13 +2,19 @@ import React from 'react';
...
+import ButtonIcon from '@/components/ui/ButtonIcon';
+import SearchOffIcon from '@/components/ui/icons/SearchOff';
 
 interface InputProps extends React.InputHTMLAttributes<HTMLInputElement> {
   className?: string;
   description?: string;
   error?: string;
   label?: string;
   required?: boolean;
+  onClear?: () => void;
```
chloe-renaud, line 19: Add a comment to explain what it is used for. It is a special use case.

Comment should be written as `/** my comment */` on the line just before to work as jsdoc (it will be displayed on props hover wherever we use it).

next/src/components/ui/ButtonIcon.tsx
```diff
@@ -9,6 +9,7 @@ interface ButtonIconProps
   title?: string;
   buttonStyle?: ButtonIconStyle;
   onClick?: () => void;
+  iconProps?: React.SVGProps<SVGSVGElement>; // Props to pass to the Icon
```
chloe-renaud: Write the comment like this:

```
/** Props to pass to the Icon */
iconProps?: React.SVGProps<SVGSVGElement>;
```

If you want JSDoc to work.

next/src/components/ui/FilterList.tsx
chloe-renaud, line 1: I'd rather have it named "Filters" but it's a matter of preference.

next/src/components/ui/FilterList.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+import ButtonIcon from '@/components/ui/ButtonIcon';
+import CloseIcon from '@/components/ui/icons/CloseIcon';
+import { Filter } from '@/models/filter';
+
+interface FilterListProps {
+  filters: Filter[];
+}
+
+const FilterList: React.FC<FilterListProps> = ({ filters = [] }) => {
```
chloe-renaud: Why not use a function like everywhere else?

You should add a comment to explain what it is used for and how it should be used.

next/src/components/codesLists/overview/CodesListsOverview.tsx
chloe-renaud, line 1: To make it easier to test (and maintain), we could create a new component, `CodesListsFilters` to handle the filters display and filtering logic.

We could pass to it as props our `codesLists` and a `setFilteredCodesLists`.

For example:

```
function MyComponentOverview({ data }) {
  const [filteredData, setFilteredData] = useState<Data>(data);

  return (
    <>
      <MyComponentFilters data={data} setFilteredData={setFilteredData} />
      <MyComponent data={filteredData} />
    </>
  )
}
```

chloe-renaud (changes requested):
Check Sonar quality gate to add coverage, remove duplication and fix issues.

PR #985, opened 2025-03-31
fix: prevent focus in collapsed codes lists

chloe-renaud:


MailineN (approved):

MailineN:
Fix tests before merge

QRuhier:
@MailineN I fixed tests, but also checked concerning accessibility : 
- using `hidden` is better since `aria-hidden` only hides for screen readers and forces to set tabIndex on all focusable content (with `hidden` it hides correctly all the content and removes focusability)
- on expand button, it's a good practice to add `aria-expanded` with the boolean if it is expanded or collapsed
- on expand button, it's a good practice to add `aria-controls` pointing to the id of the div it expands/collapses

PR #988, opened 2025-04-04
feat: enable to filter a dynamic table measure

QRuhier:
In dynamic table only (table with type = 'List'), when defining measures, add a VTL field for defining a condition of filter.

In term of model, in the response, add a new prop :
`"conditionFilter": "the condition formula i want"`

nb : we named it "conditionFilter" to be coherent with the rest of our app, even if would be more logical to call it "filterCondition"

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pogues",
-  "version": "1.8.9-rc.0",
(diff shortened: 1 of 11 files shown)


QRuhier:
see for improving UI for changing "apply filter : yes/no" for displaying the VTL condition field

MailineN (approved):

PR #992, opened 2025-04-10
fix: refresh codesLists list when duplicate/delete a codeList

QRuhier:
Refresh correctly the list of codesLists when we duplicate or delete one.

Nb: we just invalidate query, so we do not lose current filters


QRuhier:
queryKey did not have been updated when we changed the way we get the codesLists (from get questionnaire -> get codesLists) (see. https://github.com/InseeFr/Pogues/pull/982/files#diff-a69d0ee0af9c129caa7ae09769850838bdbc3a2aed1210626e889e1812f61111)

MailineN (approved):

PR #993, opened 2025-04-15
feat: made the left bar element sticky

MailineN:


QRuhier (approved):

QRuhier:
Because of the header, since the QuestionnaireNavigation has h-screen, compared to before it causes a scroll (until you do not see the header) even if there is no reason , but i guess that's ok.

PR #994, opened 2025-04-15
feat: add filter to only show unused codelist

MailineN:


next/src/components/ui/FilterList.tsx
```diff
@@ -23,25 +28,42 @@ const FilterList: React.FC<FilterListProps> = ({
 
   return (
     <>
```
QRuhier: You can remove it since now you have only one children

next/src/components/ui/FilterList.tsx
```diff
@@ -23,25 +28,42 @@ const FilterList: React.FC<FilterListProps> = ({
...
+              onClick={() => {
+                updateFilterContent(filter.filterType, !filter.filterContent);
+              }}
+              onKeyDown={(e) => {
+                if (e.key === 'Enter' || e.key === ' ') {
+                  e.preventDefault();
+                  updateFilterContent(filter.filterType, !filter.filterContent);
+                }
+              }}
+            >
```
QRuhier: It would be nice to change cursor on hover. Currently if you hover the text or the rest of the div you have different cursors and that's not obious you can click on it

QRuhier (changes requested):

PR #1011, opened 2025-06-06
fix: handle network error

chloe-renaud:
For now we simply display default "Network error" message when api request throws an error.


laurentC35 (approved):
Seems good to me.
I think some imrovment can be made (useCallBack -> codelistFilter)

But THIS works is awesome ! 🚀

PR #1029, opened 2025-06-25
fix(legacy): handle separately codesLists & nomenclatures

QRuhier:
Handle separetely codesList & nomenclatures used in questions, fixing issue : when having a UCQ with a suggester, it was impossible to update the question into a radio (or checkbox/dropdown), else the new codesList was not kept.

Also fix another issue : 
- question validation for UCQ in table (sometimes the validation did not work)

Also made some changes : 
- remove many old code concerning codesList creation inside a question (that is not possible anymore)
- change suggester form labels (now talking about nomenclature, instead of codes list)

Nb : even if we now separate codesLists & nomenclatures, in the model we still store it together as codesLists. No change done on the model


src/model/transformations/component.jsx
```diff
@@ -784,21 +719,6 @@ function storeToRemoteNested(
 
       remote.codeFilters = codeFilters;
     }
-    if (
```
chloe-renaud, line 787: Why is this removed? What did it do?

Was it to update the code list when we used a code list in a table?

src/widgets/component-new-edit/components/response-format/single/response-format-single.jsx
```diff
@@ -42,18 +35,6 @@ function ResponseFormatSingle({
     ? `${selectorPathParent}.${selectorPath}`
     : selectorPath;
 
-  useEffect(() => {
```
chloe-renaud, line 45: There is no need to have them anymore?

chloe-renaud (approved):
Great job!

src/model/transformations/component.jsx
QRuhier: It was used for get the clarification associated with the response, but in a table we don't enable to add clarification so there is no point to do this

src/widgets/component-new-edit/components/response-format/single/response-format-single.jsx
QRuhier: No i completely removed the attribute `panel` that was only used for knowing when we were created a codes list (panel = NEW) in a question.
Actually the panel = REF done here were not used in the entire app, maybe it would have been usefull for separating codeList/nomenclature, but that was not clear at all

src/model/transformations/component.jsx
chloe-renaud: I guess that's a remainder from when code lists clarification was not question-related but code list-related.

QRuhier: Yeah indeed, then it was added to every question using it, even table

PR #1030, opened 2025-06-26
fix: multiple api calls for getting questionnaire

QRuhier:
Stop getting 3 times the questionnaire from api on questionnaire page


laurentC35 (approved):
Approve but since this PR https://github.com/InseeFr/Pogues/pull/1026
Next doesn't work with Legacy in my local env. (but works before this merged)
I'ill investigate why (tansack version upgraded i think)

MailineN (approved):

chloe-renaud (approved):

PR #1033, opened 2025-06-30
feat(next): variables page

chloe-renaud:


next/src/i18n/locales/fr.json
QRuhier, line 1: Why is there "variable" and "variables" group ?

next/src/i18n/locales/fr.json
```diff
@@ -157,7 +158,39 @@
...
+      "calculated": "Calculée",
+      "collected": "Collectée",
+      "external": "Externe",
+      "label": "Type de variable"
+    }
+  },
   "variables": {
-    "title": "Variables"
+    "create": "Créer une variable",
+    "datatype": "Datatype",
```
QRuhier: It has not been translated here.
And as you use it in the filter, maybe it should be the same key as variable.datatype.label that is used in the variable table ?

next/src/i18n/locales/fr.json
chloe-renaud: Variable is for single variable page (future create, edit), Variables is for overview. The other pages follow the same pattern.

QRuhier (approved):

PR #1035, opened 2025-07-01
feat: allow to restore readonly saves

chloe-renaud:


QRuhier (approved):
Little issue with the restore confirm dialog which can be behind the "navigation bar" on questionnaire main page of Pogues-legacy. Except this, great !

PR #1038, opened 2025-07-07
feat(next): personalization page

MailineN:


next/package.json
```diff
@@ -24,11 +24,13 @@
     "@making-sense/vtl-2-0-monaco-tools-ts": "^0.1.0",
     "@tanstack/react-query": "^5.64.1",
     "@tanstack/react-router": "^1.121.34",
+    "@types/papaparse": "^5.3.16",
```
laurentC35: I think it should be a dev dependencies

next/package.json
```diff
@@ -60,6 +63,7 @@
     "globals": "^15.14.0",
     "jsdom": "^26.0.0",
     "nock": "^14.0.1",
+    "papaparse": "^5.5.3",
```
chloe-renaud: Ca devrait être dans dependencies, pas devDepdencies.

next/src/routes/_layout/questionnaire.$questionnaireId/_layout-q/personalize/$publicEnemyId.tsx
```diff
@@ -0,0 +1,87 @@
...
+import ContentHeader from '@/components/layout/ContentHeader';
+import ContentMain from '@/components/layout/ContentMain';
+import EditPersonalization from '@/components/personalization/form/edit/EditPersonalization';
+import { PersonalizationQuestionnaire } from '@/models/personalizationQuestionnaire';
+
+/**
+ * Page that allow to edit survey unit dataset.
+ */
+export const Route = createFileRoute(
+  '/_layout/questionnaire/$questionnaireId/_layout-q/personalize/$publicEnemyId',
```
chloe-renaud: Shouldn't the route be `personalization` instead of `personalize`?

next/src/routes/_layout/questionnaire.$questionnaireId/_layout-q/personalize/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  children,
+}: Readonly<{
+  children: React.ReactNode;
+}>) {
+  const { t } = useTranslation();
+  return (
+    <>
+      <ContentHeader
+        title={`${t('personalization.overview.title')}`}
+        action={null}
```
chloe-renaud: No need to specify `action` if it's null

next/src/models/questionnaires.ts
```diff
@@ -11,10 +11,10 @@ export type Questionnaire = {
 };
 
 export enum TargetModes {
-  CAWI,
-  CAPI,
-  PAPI,
-  CATI,
+  CAWI = 'CAWI',
```
chloe-renaud: No need to add a value to the enum, it works well as is and is more optimized.

next/src/routes/_layout/questionnaire.$questionnaireId/_layout-q/personalize/$publicEnemyId.tsx
```diff
@@ -0,0 +1,87 @@
...
+)({
+  component: RouteComponent,
+  errorComponent: ({ error }) => <ErrorComponent error={error} />,
+  loader: async ({
+    context: { queryClient },
+    params: { questionnaireId, publicEnemyId },
+  }) => {
+    await queryClient.ensureQueryData(
```
chloe-renaud: We should put that in a promise that will encompass both promises. That way we don't have to wait for both promises to conclude one after the other and we can run both of them at the same time.

See this for example: https://github.com/TanStack/query/discussions/5553#discussioncomment-6143015

next/src/routes/_layout/questionnaire.$questionnaireId/_layout-q/personalize/$publicEnemyId.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const { data: csvData } = useSuspenseQuery(
+    personalizationFileQueryOptions(publicEnemyId),
+  );
+
+  const [parsedCsv, setParsedCsv] = useState<ParseResult<unknown> | null>(null);
+
+  useMemo(() => {
+    if (!csvData) return;
+
+    csvData.text().then((csvText) => {
```
chloe-renaud: I think it would be better to put that directly in our query.

i.e. whenever we fetch the csv file, parse it and simply return the parsed data (in /api functions). We won't need to handle the weird csv data in our page components that way.

next/src/models/personalizationQuestionnaire.ts
```diff
@@ -0,0 +1,53 @@
+import { TargetModes } from '@/models/questionnaires';
+
+export type PersonalizationQuestionnaire = {
+  id: string;
+  poguesId: string;
+  label: string;
+  modes: Modes[];
+  context: SurveyContext;
+  surveyUnitData: File | undefined;
```
chloe-renaud: If it can be undefined, use `?` instead.

next/src/models/personalizationQuestionnaire.ts
```diff
@@ -0,0 +1,53 @@
+import { TargetModes } from '@/models/questionnaires';
+
+export type PersonalizationQuestionnaire = {
+  id: string;
+  poguesId: string;
+  label: string;
+  modes: Modes[];
```
chloe-renaud: Should be `Mode` singular (it becomes "modes" because it is a `[]`).

next/src/models/personalizationQuestionnaire.ts
```diff
@@ -0,0 +1,53 @@
...
+  context: SurveyContext;
```
chloe-renaud, line 6: What's this ? Household and Business ? We should pass an enum and handle the transformation in /api for simplicity sake. We don't need this weird type in our codebase.

next/src/models/personalizationQuestionnaire.ts
```diff
@@ -0,0 +1,53 @@
...
+  isSynchronized: boolean;
+};
+
+export type Modes = {
+  name: TargetModes;
+  isWebMode: boolean;
+};
+
+export type SurveyContext = {
+  name: string;
```
chloe-renaud: If we really want to keep it that way the string should be defined as which value it can truly be.

next/src/i18n/locales/fr.json
```diff
@@ -143,11 +143,53 @@
...
     "title": "Questionnaire"
   },
+  "personalization": {
+    "create": {
+      "business": "Entreprise",
+      "context": "Contexte",
+      "createQuestionnaire": "Charger les données dans les orchestrateurs",
+      "createQuestionnaireDescription": "Créer le questionnaire dans les orchestrateurs avec les données chargées.",
+      "description": "Créez une personnalisation pour pré-remplir le questionnaire avec les données des unités enquêtées.",
+      "download_error": "Erreur lors du téléchargement du schéma CSV",
```
chloe-renaud: Don't use `_` in translation keys.

next/src/components/ui/ExpandButton.tsx
```diff
@@ -17,11 +17,12 @@ export default function ExpandButton({
   const { t } = useTranslation();
   return (
     <button
-      className="cursor-pointer"
+      className="cursor-pointer mt-2"
       onClick={() => setIsExpanded((v) => !v)}
       aria-expanded={isExpanded}
       aria-controls={ariaControls}
       aria-label={isExpanded ? t('common.collapse') : t('common.expand')}
+      style={{ marginTop: '0.5rem' }}
```
chloe-renaud: Why is the style defined twice? Besides you shouldn't change it here, it is used by OverviewItem.

next/src/components/ui/ButtonLink.tsx
```diff
@@ -9,27 +9,55 @@ export enum ButtonStyle {
...
+      buttonStyle = ButtonStyle.Secondary,
+      disabled = false,
+      children,
+      ...props
+    },
+    ref,
+  ) => {
+    let buttonClass = '';
+    if (buttonStyle === ButtonStyle.Primary) {
+      buttonClass = disabled
```
chloe-renaud: Please use disabled:my-class instead.

next/src/components/personalization/form/CsvViewerTable.tsx
```diff
@@ -0,0 +1,50 @@
...
+/** Display uploaded csv file as a table. */
+export default function CsvViewerTable({
+  parsedCsv,
+}: Readonly<CsvViewerTableProps>) {
+  const shouldScroll = parsedCsv.data.length > 4;
+  return (
+    <div className="overflow-x-auto w-full my-3">
+      <div
+        style={{
+          maxHeight: shouldScroll ? '320px' : 'none',
```
chloe-renaud: Can't we play around a `max-h` instead?

chloe-renaud (changes requested):

next/src/models/personalizationQuestionnaire.ts
MailineN: The back office gives and needs this pattern for context :` {"name":"BUSINESS","value":"Entreprise"}`
I've switched the type to 
```
export type SurveyContext = {
  name: 'BUSINESS' | 'HOUSEHOLD';
  value?: 'Entreprise' | 'Ménage';
};
```

next/src/models/questionnaires.ts
MailineN: This works correctly when displaying the mode on the page, but in tests, the enum is treated as a numeric enum (e.g., TargetMode.CAWI = 0), which causes the test to fail.

I’d gladly welcome suggestions that avoid adding too much logic to my text component.

next/src/components/ui/ExpandButton.tsx
MailineN: I’ve moved it into className. It’s set to prevent the button from overlapping with the element above it — which was bothering me.
The issue was also present in OverviewItem.

next/src/components/personalization/form/PersonalizationForm.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+      name: SurveyContextEnum.HOUSEHOLD,
+      value: SurveyContextValueEnum.HOUSEHOLD,
+    },
+    surveyUnitData: undefined,
+    isSynchronized: true,
+  };
+
+  const setQuestionnaire = vi.fn();
+  const setErrorUpload = vi.fn();
+  const handleSubmit = vi.fn();
```
chloe-renaud: If it's never used, we can simply declare the mock function when we render the component, no need to make a `const` out of it which will never be used. 

```
<PersonalizationForm
  handleSubmit={vi.fn()}
/>
```

next/src/i18n/locales/en.json
```diff
@@ -122,6 +122,47 @@
...
+      "description": "Upload survey units to pre-fill the questionnaire.",
+      "downloadError": "Error downloading CSV schema",
+      "downloadSuccess": "CSV schema downloaded successfully",
+      "error": "Error creating personalization",
+      "fileViewerError": "Error displaying the file.",
+      "household": "Household",
+      "noFileChosen": "No file chosen",
+      "saveError": "Error saving questionnaire",
+      "saveSuccess": "Questionnaire saved successfully",
+      "schema": "Expected file schema",
```
chloe-renaud: It would be better if the key was `create.expectedSchema` (or `create.expectedFileSchema`). It should be more obvious what the translation is about when we are reading the various app components.

For example, at first glance I thought `create.schema` was about creating a schema and I was confused.

next/src/components/personalization/overview/PersonalizationTile.tsx
```diff
@@ -0,0 +1,31 @@
...
+  children,
+}: Readonly<PersonalisationTileProps>) {
+  const { t } = useTranslation();
+
+  return (
+    <div className="relative bg-default p-4 border border-default shadow-md grid grid-rows-[auto_1fr_auto]">
+      <div className="flex flex-col ">
+        <h3>{data.label}</h3>
+        <span className="text-sm text-gray-600 mt-1">
+          {t('personalization.overview.modes')}:{' '}
```
chloe-renaud: The `:` should be in the translation as it's `something:` in english and `something :` in french.

QRuhier (approved):

PR #1040, opened 2025-07-08
feat(next): more generic filters

chloe-renaud:
I removed the useFilters hook and use a Filters component instead which handle the display the various filter inputs and the filtering.

Instead of saying "I want a 'name' and 'not used' filters", we say "I want a text filter whose label is 'name', and a boolean filter whose label is 'not used'" which should allow for more flexibility in the long run.

The design has changed a bit. Feel free to give me feedbacks.


next/src/components/codesLists/overview/CodesListsOverview.tsx
```diff
@@ -26,52 +26,37 @@ export default function CodesListsOverview({
...
+      label: t('codesLists.notUsed'),
+      onFilter: (v: CodesList, filter?: string) =>
+        filter === 'notUsed'
+          ? !v.relatedQuestionNames || v.relatedQuestionNames.length === 0
+          : true,
+      options: [
+        { label: t('codesLists.all'), value: 'all' },
+        { label: t('codesLists.notUsed'), value: 'notUsed' },
+      ],
+      type: FilterType.ToggleGroup,
```
MailineN, line 40: A switch-type filter (or alternatively a default value) would be clearer for most users. 
Currently, when clicking again on “Unused,” the filter is correctly removed, but there’s no clear indication that all code lists are now being displayed. Same goes for "All.”
This could be confusing if there are many code lists, even though the total number of code lists is shown on the right.

QRuhier (approved):
Nice, it'll be easier for handling other filters in the future !

next/src/components/codesLists/overview/CodesListsOverview.tsx
chloe-renaud: I wanted to use a switch but it wasn't very pretty.

I did not notice the weird behaviour when we clicked on "not used" again and "all" was not checked though.

Maybe we could put a switch-like button instead?

MailineN: That was the case before, a sort of pill-shaped button that acted as a switch.
I tried using a proper switch filter instead, but the UI wasn’t consistent with the rest of the page (and looked ugly in my opinion).
If we really want to use the Switch component, it would requires some heavy stylisation

MailineN (approved):

PR #1041, opened 2025-07-09
feat(legacy): remove personalization link in old editor

MailineN:


QRuhier (approved):

MailineN:
À merger quand #1038 sera prêt

chloe-renaud (changes requested):
This is not a "refactor" commit, refactor is for when a change in the code base does not change features (performance improvement, or when you change a weird function to make it clearer - but the input and output are the same). This is more of a feat commit.

PR #1056, opened 2025-07-22
feat: switch to interrogations

MailineN:


QRuhier:
@MailineN It would be nice to rename surveyUnit to interrogation everywhere for coherency, but we can do it later

QRuhier (approved):

PR #1068, opened 2025-07-31
feat(next): pre collected interrogations

MailineN:
(Had to start over because I messed up with a rebase)


next/src/api/personalization.ts
```diff
@@ -16,65 +15,49 @@ import { getFileName, openDocument } from './utils/personalization';
...
+export const basePersonalizationQueryOptions = (poguesId: string) => ({
+  queryKey: ['personalization', 'base', poguesId],
+  queryFn: () => getPublicEnemyBaseData(poguesId),
 });
+
 export async function getPublicEnemyBaseData(
-  questionnaireId: string,
-): Promise<[PersonalizationQuestionnaire, ParseResult]> {
+  poguesId: string,
+): Promise<[InterrogationModeDataResponse, ParseResult | string]> {
```
chloe-renaud: TS is unhappy that `ParseResult<T>` does not have its type argument T. Worst case scenario you can use `ParseResult<unknown>`.

next/src/api/personalization.ts
```diff
@@ -16,65 +15,49 @@ import { getFileName, openDocument } from './utils/personalization';
  *
  * @see {@link getPublicEnemyBaseData}
  */
-export const basePersonalizationQueryOptions = (questionnaireId: string) => ({
-  queryKey: ['personalization', 'base', questionnaireId],
-  queryFn: () => getPublicEnemyBaseData(questionnaireId),
+export const basePersonalizationQueryOptions = (poguesId: string) => ({
+  queryKey: ['personalization', 'base', poguesId],
```
chloe-renaud: We should declare our keys in a `personalizationKeys` const. This way, we can easily set queries as invalid through this `personalizationKeys` without any risk for mistake later on, if a key is changed or there is a typo.

(You can check how it is done in other api files.)

next/src/api/personalization.ts
```diff
@@ -16,65 +15,49 @@ import { getFileName, openDocument } from './utils/personalization';
...
 export async function getPublicEnemyBaseData(
-  questionnaireId: string,
-): Promise<[PersonalizationQuestionnaire, ParseResult]> {
+  poguesId: string,
+): Promise<[InterrogationModeDataResponse, ParseResult | string]> {
   try {
-    const existingData = await getExistingPublicEnemyData(questionnaireId);
-    const fileData = await getExistingCsvSchema(existingData.id);
-    return [existingData, fileData];
+    const interrogations: InterrogationModeDataResponse =
```
chloe-renaud: If we can trigger the promises at the same time, it is a better practice to use `Promise.all` to improve performance.

See MDN documentation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all

next/src/api/personalization.ts
chloe-renaud, line 1: We should keep calling the param `questionnaireId` instead of `poguesId`, it would be clearer I think.

next/src/components/personalization/form/PersonalizationForm.test.tsx
```diff
@@ -37,6 +35,10 @@ vi.mock('react-hot-toast', () => ({
   default: { error: vi.fn(), success: vi.fn(), promise: vi.fn() },
 }));
 
+vi.mock('@/api/personalization', () => ({
```
chloe-renaud: It shouldn't be mocked twice.

next/src/components/personalization/form/PersonalizationForm.tsx
```diff
@@ -67,41 +71,52 @@ export default function PersonalizationForm({
...
+      return checkInterrogationsData(questionnaireId, file).then((response) => {
+        setUploadMessage(response as UploadMessage);
+        setIsErrorUpload(false);
+      });
     },
     onError: (error: AxiosError) => {
       toast.error(t('personalization.create.uploadError'));
-      setErrorUpload(error.response?.data as UploadError);
+      setUploadMessage(error.response?.data as UploadMessage);
+      console.log('File data checked with error:', error);
```
chloe-renaud:
```suggestion
      console.error('File data checked with error:', error);
```

chloe-renaud (approved):

PR #1069, opened 2025-08-01
feat(next): add articulation pages

QRuhier:
Handle articulation on a questionnaire.

Currently what has been done :  
- main articulation page
  - if there is no variable : display error saying questionnaire does not have roundabout
  - if there is no articulation : display a create button for creating articulation
  - if there are articulation and variables : display existing articulation, with possibility of editing or deleting => delete will remove articulation object, so after refresh we are in the previous case
- create/edit articulation page
  - 3 fields with the articulation items (name, gender, name), that are all mandatory


QRuhier:
Waiting for new endpoints on pogues-api to handle articulation : 
- create articulation
- edit articulation
- delete articulation
- get articulation
- get variables having scope being the loop on which the roundabout is based

QRuhier:
Don't hesitate to improve UI.

For overview, i made a table since in the end in application using articulation it will be this table with one raw per iteration (interpreted formula for each)

QRuhier:
Todo : since articulation is only for VTL questionnaire, see how handle xpath questionnaire : 
- error on articulation page ? -> handled directly by api when getting articulation ?
- disable articulation button in lateral menu ? -> would be nice to have a new endpoint giving the language for a questionnaireId

QRuhier:
nb: do a PR on Pogues-legacy to handle articulation object in state (for keeping it untouched between the get questionnaire and put questionnaire)

MailineN:
Instead of a table with only one row, we could display the articulation formula as a completed form in a single column after its creation. This also makes it clearer to the user that these formulas can be modified if needed.

next/src/api/articulation.ts
```diff
@@ -0,0 +1,135 @@
...
+  });
+
+/**
+ * Used to retrieve articulation variables associated to a questionnaire.
+ *
+ * @see {@link getArticulationVariables}
+ */
+export const articulationVariablesQueryOptions = (questionnaireId: string) =>
+  queryOptions({
+    queryKey: ['variables', questionnaireId] as const,
```
chloe-renaud: The key should be related to articulation variables, you'll have an error between variables pages and articulations page otherwise (besides it should be defined in the articulationKeys const).

next/src/i18n/locales/en.json
```diff
@@ -233,5 +234,40 @@
...
+      "firstName": "First Name",
+      "gender": "Gender",
+      "age": "Age"
+    },
+    "overview": {
+      "variableTable": "Articulation table variables",
+      "title": "Articulation",
+      "noArticulation": "No articulation is specified",
+      "error": {
+        "formulaNotVtl": "The questionnaire needs to use the formula language VTL",
```
MailineN: "The questionnaire must use the VTL formula language"

next/src/i18n/locales/en.json
```diff
@@ -233,5 +234,40 @@
...
+      }
+    },
+    "create": {
+      "create": "Create an articulation",
+      "success": "The articulation has been created",
+      "title": "New articulation"
+    },
+    "delete": {
+      "dialogTitle": "Delete articulation",
+      "dialogConfirm": "Are you sure you want to delete articulation?",
```
MailineN: Add "the" as it is marked above

MailineN (approved):

next/src/models/articulation.ts
```diff
@@ -0,0 +1,20 @@
+// for now we handle precisely 3 items with fixed label
+export type ArticulationItems = [
+  { label: 'Prénom'; value: string },
+  { label: 'Sexe'; value: string },
+  { label: 'Age'; value: string },
+];
+
+export type Articulation = { items: ArticulationItems };
+
+export const ARTICULATION_ITEMS_TRANSLATIONS = {
```
chloe-renaud, line 10: What is this

Since we are locked to the three labels "Prénom", "Sexe", "Age", maybe we should be explicit and not translate them.

next/src/i18n/locales/fr.json
```diff
@@ -233,5 +234,40 @@
...
+      "firstName": "Prénom",
+      "gender": "Sexe",
+      "age": "Âge"
+    },
+    "overview": {
+      "variableTable": "Variables du tableau d'articulation",
+      "title": "Articulation",
+      "noArticulation": "Aucune articulation n'a été définie",
+      "error": {
+        "formulaNotVtl": "Le questionnaire doit utiliser le language VTL",
```
chloe-renaud: Peut-être qu'on peut rendre l'erreur plus explicite tel que :

> "L'articulation n'est disponible que pour les formulaires en VTL."
> "L'articulation n'est disponible que pour les formulaires avec un rond-point."

chloe-renaud (changes requested):
Apart from the key thing, it looks good for now. Since it is hidden by default we can improve the design in another PR before official delivery if needed.

chloe-renaud (approved):

PR #1094, opened 2025-09-26
feat(next): enable to delete multimode

QRuhier:
In multimode page, add a "delete" button.

Currently if the user wants to remove its multimode rules, he needs to edit them cleaning every field.
We prefer to add this shortcut, which is also more coherent with what is done on articulation

### DIFF
--- a/next/src/components/multimode/MultimodeOverview.test.tsx
+++ b/next/src/components/multimode/MultimodeOverview.test.tsx
@@ -1,18 +1,11 @@
 import { fireEvent, waitFor } from '@testing-library/dom';
+import nock from 'nock';
 
-import { deleteMultimode } from '@/api/multimode';
 import { renderWithRouter } from '@/testing/render';
 
 import MultimodeOverview from './MultimodeOverview';
 
-// mock the API function
-vi.mock('@/api/multimode', async (importOriginal) => {
-  const actual = await importOriginal<typeof import('@/api/multimode')>();
-  return {
-    ...actual,
-    deleteMultimode: vi.fn(),
-  };
-});
+vi.mock('@/lib/auth/oidc');
 
 describe('MultimodeOverview', () => {
   it('display multimode questionnaire rule', async () => {
@@ -98,14 +91,18 @@ describe('MultimodeOverview', () => {
 
     expect(getByRole('button', { name: 'Delete' })).toBeEnabled();
 
+    const scope = nock('https://mock-api')
+      .delete('/persistence/questionnaire/q-id/multimode')
+      .reply(200);
+
     // click on delete button
     fireEvent.click(getByRole('button', { name: 'Delete' }));
     // confirm in dialog
     fireEvent.click(getByRole('button', { name: 'Validate' }));
 
+    // multimode has been deleted
     await waitFor(() => {
-      // multimode has been deleted
-      expect(deleteMultimode).toHaveBeenCalledWith('q-id');
+      expect(scope.isDone()).toBeTruthy();
     });
   });
 
--- a/next/src/components/multimode/MultimodeOverview.test.tsx
+++ b/next/src/components/multimode/MultimodeOverview.test.tsx
@@ -1,6 +1,18 @@
+import { fireEvent, waitFor } from '@testing-library/dom';
+
+import { deleteMultimode } from '@/api/multimode';
 import { renderWithRouter } from '@/testing/render';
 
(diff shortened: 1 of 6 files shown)


next/src/components/multimode/MultimodeOverview.test.tsx
```diff
@@ -1,7 +1,19 @@
+import { fireEvent, waitFor } from '@testing-library/dom';
+
+import { deleteMultimode } from '@/api/multimode';
 import { renderWithRouter } from '@/testing/render';
 
 import MultimodeOverview from './MultimodeOverview';
 
+// mock the API function
+vi.mock('@/api/multimode', async (importOriginal) => {
```
chloe-renaud: I personally like mocking the API call more (see ReadonlyWarning test).

i.e.

```
    const scope = nock('https://mock-api')
      .post('/persistence/questionnaire/restore/my-version')
      .reply(200);

    await user.click(getByRole('button', { name: /Restore/i }));

    expect(
      getByText(/A new save will be created with this save data./i),
    ).toBeInTheDocument();

    await user.click(getByRole('button', { name: /Validate/i }));

    expect(scope.isDone()).toBeTruthy();
```

That way, if we ever change how the API calls are implemented, we don't have to change this test. And the test is more "what happen to the user" than "what happen in our codebase".

next/src/components/multimode/MultimodeOverview.tsx
```diff
@@ -1,9 +1,14 @@
+import { useMutation, useQueryClient } from '@tanstack/react-query';
+import toast from 'react-hot-toast';
 import { useTranslation } from 'react-i18next';
 
+import { deleteMultimode, multimodeKeys } from '@/api/multimode';
 import ButtonLink from '@/components/ui/ButtonLink';
 import Card from '@/components/ui/Card';
 import type { MultimodeIsMovedRules } from '@/models/multimode';
 
+import DialogButton from '../ui/DialogButton';
```
chloe-renaud:
```suggestion
import DialogButton from '@/components/ui/DialogButton';
```

next/src/components/multimode/MultimodeOverview.tsx
```diff
@@ -61,12 +90,20 @@ export default function MultimodeOverview({
             <div>{t('multimode.noLeafRule')}</div>
           )}
           {readonly ? null : (
-            <ButtonLink
-              to="/questionnaire/$questionnaireId/multimode/edit"
-              params={{ questionnaireId }}
-            >
-              {t('common.edit')}
-            </ButtonLink>
+            <div className="flex gap-x-2">
```
chloe-renaud, line 92: I think it would have more sense to have a "delete multimode" button at the bottom of the page (with the text / border in red) than in the card. That way it won't be next to the "edit" (which is a bit confusing in this case imo) and the action of "i want to delete everything in this page" will be more obvious.

next/src/lib/i18n/locales/fr/multimode.json
```diff
@@ -1,4 +1,9 @@
 {
+  "delete": {
```
chloe-renaud, line 2: Si tu es ok avec mon suggestion plus haut je verrais un message de suppression plus global du style :

"Supprimer le multimode"
"Êtes-vous sûr de vouloir supprimer la fonctionnalité du multimode ?"
"Le multimode a été supprimé"

chloe-renaud (commented):
Looks good.

I have two suggestions for the test and delete button, but you can disagree and that's not mandatory if this change is urgent.

next/src/components/multimode/MultimodeOverview.tsx
QRuhier: But the edit button is also for editing everything, so the delete button applies on the same content. And it's kinda strange to separate those 2 buttons since on this page you can only act on the multimode rules.

Maybe if in the future there are more possibilities on this page yeah we could differenciate multimode =/= rules

chloe-renaud: It's more like I'm not a fan of having the "delete" button next to the "edit" one

It seems like something you would not normally do (so it should not be in the middle of the screen) and the UX should make it more explicit that you delete the whole feature.

It would allow to put it in red without it being weird (by taking too much space next to edit button).

But I can make a separate PR to test it out so we can experiment both versions and see which one works best though.

chloe-renaud (approved):

PR #1095, opened 2025-09-26
feat: handle dirty state in existing forms

QRuhier:
Handle dirty state in existing forms (articulation, codes list, multimode) and in legacy component for questionnaire page.

Add a context for having a global dirty state state in the app (default : false).
Forms update this state continuously depending if the form is dirty.

When the user tries to navigate to any page while dirty state, a custom confirm dialog is displayed. The user can cancel, or confirm then the dirty state is reset and we navigate.

When the user tries to close tab while dirty state, the default navigator confirm dialog is displayed.


next/src/contexts/dirtyStateContext.test.tsx
chloe-renaud, line 1: File name begins with a lowercase.

next/src/contexts/DirtyStateContext.tsx
```diff
@@ -0,0 +1,65 @@
...
+ */
+export function DirtyStateProvider({
+  children,
+}: Readonly<{
+  children: ReactNode;
+}>) {
+  const [isDirty, setIsDirty] = useState(false);
+
+  /** Mark the app as dirty/clean */
+  const setDirty = useCallback((dirty: boolean) => {
```
chloe-renaud: Why don't we directly hand `setIsDirty` over?

Besides I'm not a fan of the `setDirty` method which is a bit confusing (the name implies it should be called without any param and it will set the state as dirty). I'd rather have `setIsDirtyState` which is more explicit (or two methods `setDirtyState` and `removeDirtyState`).

next/src/contexts/dirtyStateContext.test.tsx
```diff
@@ -0,0 +1,37 @@
+import { act, renderHook } from '@testing-library/react';
+
+import {
+  DirtyStateProvider,
+  useDirtyState,
+} from '@/contexts/DirtyStateContext';
+
+describe('DirtyStateContext', () => {
+  it('should have default isDirty=false', () => {
```
chloe-renaud:
```suggestion
  it('should not be in dirty state by default', () => {
```

next/src/components/articulation/create/CreateArticulation.tsx
```diff
@@ -46,11 +49,16 @@ export default function CreateArticulation({
     const promise = mutation.mutateAsync(
       { articulation, questionnaireId },
       {
-        onSuccess: () =>
+        onSuccess: () => {
+          // flushSync forces the state to update immediately, before trying to navigate
+          flushSync(() => {
```
chloe-renaud: Do we really need this? It seems to work fine without it.

next/src/components/dirtyState/DirtyStateDialog.tsx
```diff
@@ -0,0 +1,46 @@
...
+    shouldBlockFn: () => isDirty,
+    withResolver: true,
+    enableBeforeUnload: isDirty,
+  });
+
+  useEffect(() => {
+    if (blocker.status === 'blocked') {
+      setDialogOpen(true);
+    }
+  }, [blocker.next, blocker.status]);
```
chloe-renaud: Why is `blocker.next` in dependency array?

next/src/components/dirtyState/DirtyStateDialog.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { useBlocker } from '@tanstack/react-router';
+import { useTranslation } from 'react-i18next';
+
+import Dialog from '@/components/ui/Dialog';
+import { useDirtyState } from '@/contexts/DirtyStateContext';
+
+export function DirtyStateDialog() {
+  const { t } = useTranslation();
+  const { isDirty, setDirty } = useDirtyState();
+  const [dialogOpen, setDialogOpen] = useState(false);
```
chloe-renaud:
```suggestion
  const [isDialogOpen, setIsDialogOpen] = useState(false);
```

next/src/lib/i18n/locales/fr/common.json
```diff
@@ -25,6 +25,7 @@
   "questionnaires": "Questionnaires",
   "title": "Titre",
   "unsavedModification": "Vous n'avez pas enregistré vos modifications. Êtes-vous sûr de vouloir quitter ?",
```
chloe-renaud:
```suggestion
  "unsavedModifications": {
    "body": "Vous n'avez pas enregistré vos modifications. Êtes-vous sûr de vouloir quitter ?",
    "title": "Modifications non enregistrées"
  },
```

chloe-renaud (changes requested):
I am not a fan of using `navigate` everywhere instead of explicitely setting the href. It seems less aria-friendly too.

Can't we keep the code simpler by still using our ButtonLink, but simply use the `useBlocker` hook and use the native way of returning `true` / `false` with the default confirm thingy to handle the dirty state? (like we did in legacy index)

next/src/components/articulation/create/CreateArticulation.tsx
QRuhier: Yep it does not work without it (it tries to navigates before the state is update, so the dialog is displayed)

next/src/components/dirtyState/DirtyStateDialog.tsx
QRuhier: Because we need the dialog to open only when the user tries to navigate. 
If he cancels the dialog, the status is still blocked so if he tries again to navigate the dialog won't open. Using blocker.next as dependency ensures that the dialog reopens

next/src/components/articulation/form/ArticulationForm.tsx
```diff
@@ -1,11 +1,12 @@
 import { zodResolver } from '@hookform/resolvers/zod';
+import { useBlocker, useNavigate } from '@tanstack/react-router';
 import i18next from 'i18next';
 import { Controller, type SubmitHandler, useForm } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
 import { z } from 'zod';
 
-import Button from '@/components/ui/Button';
-import ButtonLink from '@/components/ui/ButtonLink';
+import { DirtyStateDialog } from '@/components/layout/DirtyStateDialog';
```
chloe-renaud: Should be default import to be coherent with the rest of the app.

next/src/components/articulation/form/ArticulationForm.tsx
```diff
@@ -90,16 +105,21 @@ export default function ArticulationForm({
...
+        <Button
+          type="submit"
+          disabled={!isDirty || !isValid}
+          buttonStyle={ButtonStyle.Primary}
+        >
           {t('common.validate')}
         </Button>
       </div>
+
+      {status === 'blocked' ? (
```
chloe-renaud, line 121: Maybe it would be better to put it outside of the `<form>` to avoid any issue with buttons and form interaction (shouldn't be a problem there since the Dialog is put in a React portal thingy I think but we never know how our Dialog component will evolve).

e.g.

```
<>
  <form> {/* ... */} </form>
  <DirtyStateDialog />
</>
```

next/src/components/articulation/form/ArticulationForm.test.tsx
```diff
@@ -4,9 +4,19 @@ import { renderWithRouter } from '@/testing/render';
 
 import ArticulationForm from './ArticulationForm';
 
-// mock VTLEditor as a classic input
+const mockNavigate = vi.fn();
+
 vi.mock('@/components/ui/form/VTLEditor');
 
+// Mock useNavigate from @tanstack/react-router
+vi.mock('@tanstack/react-router', async () => {
```
chloe-renaud, line 12: Not a fan but unfortunately we don't have another way to test this outside of using E2E test (which unfortunately don't exist as of now).

next/src/components/articulation/edit/EditArticulation.tsx
```diff
@@ -46,7 +46,7 @@ export default function EditArticulation({
       { articulation, questionnaireId },
       {
         onSuccess: () =>
-          navigate({
+          void navigate({
```
chloe-renaud: Sonar does not seem to be happy with this (and the IDE is not happy when it is removed, we can never win...)

next/src/components/layout/DirtyStateDialog.test.tsx
```diff
@@ -0,0 +1,45 @@
+import { fireEvent } from '@testing-library/react';
+import { vi } from 'vitest';
```
chloe-renaud: Vitest imports should be global, no need to declare it.

next/src/components/legacy/index.tsx
```diff
@@ -1,44 +1,46 @@
...
+    withResolver: true,
   });
 
   const myComponent = useMemo(
     () => legacyApp(setIsDirtyState, t),
     [setIsDirtyState, t],
   );
 
-  return myComponent;
+  return (
```
chloe-renaud, line 27: Hm... Make sure legacy navigation (e.g. accessing TCM page and such) still work if you don't return the component directly. I had an issue and introduced a bug when I tried to wrap the legacy component a few months ago.

Worst case, maybe it should be better to keep the native modal for now and explain we can remove it only when the form is implemented in our new code base and we can get rid of federation.

next/src/components/multimode/form/IsMovedRulesForm.test.tsx
```diff
@@ -74,4 +85,27 @@ describe('IsMovedRulesForm', () => {
       ),
     ).toBeInTheDocument();
   });
+
+  it('should enable cancel button, going back to multimode page', async () => {
```
chloe-renaud: No need to edit it now but I think it is better to have a more functional description. Here it is a tad technical and simply describe the test instead of the functionality (since you speak of the cancel button being "enabled"). It would make sense if it could be disabled (like the validate button which become clickable under some specific conditions) but that's not the case so it's just confusing.

e.g. "allows to go back to multimode overview page by clicking 'cancel' button".

next/src/components/ui/Dialog.tsx
```diff
@@ -8,6 +8,8 @@ import Button, { ButtonStyle } from './Button';
 interface DialogButtonProps {
   /** Body message in the dialog. */
   body: React.ReactNode;
+  /** Function to execute if the user click on "close". */
```
chloe-renaud: He actually clicks on "cancel".

next/src/components/ui/Dialog.tsx
```diff
@@ -54,7 +57,13 @@ export default function Dialog({
             {body}
           </UIDialog.Description>
           <div className="flex justify-end gap-4">
-            <UIDialog.Close render={<Button>{t('common.cancel')}</Button>} />
+            <UIDialog.Close
```
chloe-renaud: I feel like it is a bit hacky.

Maybe it should be something like that instead:

```
            {onCancel ? (
              <Button onClick={onCancel}>{t('common.cancel')}</Button>
            ) : (
              <UIDialog.Close render={<Button>{t('common.cancel')}</Button>} />
            )}
```

That way we keep the native Dialog.Close functionality and simply replace it entirely when we have a custom action to trigger on cancel (just like we do with the validate).

```
  /**
   * Function to execute if the user click on "cancel".
   * 
   * It will replace the default "cancel" button. It should only be used with a controlled dialog.
   */
  onCancel?: () => void;
```

chloe-renaud (changes requested):

next/src/components/articulation/edit/EditArticulation.tsx
QRuhier: lol yeah i did add it because i saw it on the IDE. I'll remove it

next/src/components/legacy/index.tsx
QRuhier: Yes i checked that with both apps (even in pages like TCM and questionnaire merge) and it's still ok 👍 
The problem was that we needed the useMemo for the legacy component, so we still do it

chloe-renaud (approved):

PR #1098, opened 2025-10-23
feat(next): protocoles submenu

chloe-renaud:
- Add a "protocoles" submenu with "multimode" and "articulation".
- Rename "multimode" and "articulation" pages as "Changement de mode" and "Récapitulatif de rond-point".

There might be some typo, feel free to update as needed.

I was also wondering whether we should change the uri path from "/multimode" and "/articulation"


next/pnpm-lock.yaml
QRuhier, line 1: to remove until we do the entire migration to pnpm (including CI changes)

QRuhier:
That's ok for me, but 2 functional things : 
- the "protocoles" menu should not be expanded by default (since the idea is to reduce the size of all the items in te sidebar) => warning if we reduce it by default, what do we do when we are on an articulation/multimode page, do we expand it by default for seing in the menu the current page ?
- on a small screen, currently the "protocoles" submenu opens directly when you hover the icon (no need to click), not really fan but why not. But when you select one of the sub-item, you navigate into it and the submenu is still open until you move your cursor away from it

laurentC35 (approved):

laurentC35 (changes requested):
Revoir la traduction en anglais, le mot articulation et multimode doivent plutôt être remplacé par "Récapitulatif de rond point" et "Changement de mode" (dans leur équivalent anglais bien entendu)

QRuhier (approved):

PR #1103, opened 2025-11-13
feat: disable questionnaire duplication on dirty state

MailineN:


chloe-renaud (approved):

QRuhier (commented):
Can you easily add a new test in list-with-input-panel checking it's disabled when dirty state ?

PR #1104, opened 2025-11-14
feat(next): external variable create form

chloe-renaud:


next/src/components/variables/form/VariableForm.tsx
```diff
@@ -193,56 +158,102 @@ export default function VariableForm({
...
+      ) : null}
       {selectedTypeName === DatatypeType.Numeric ? (
         <>
           <Controller
             name="datatype.minimum"
             control={control}
+            rules={{ required: true }}
             render={({ field, fieldState: { error } }) => (
+              /* @ts-expect-error minimum is a number as datatype is numeric */
               <Input
```
MailineN, line 229: This allows text input without any indication that the field requires a number. The user can’t submit the variable, but may not understand why.
The issue doesn’t occur in the edit form.

Also, the min can be greater than the max

next/src/components/variables/overview/VariableLine.test.tsx
```diff
@@ -1,21 +1,26 @@
...
-        variable={{
-          id: 'my-id',
-          name: 'MY_VAR',
-          description: 'This var likes strawberries',
-          datatype: { typeName: DatatypeType.Text },
-          type: VariableType.Collected,
-        }}
-      />,
+    const { getByText } = await renderWithRouter(
+      <table>
```
MailineN, line 10: Why do you need this ?

chloe-renaud: Vitest displays a bunch of error because a td must be inside of a table, so it reduces useless errors / logs.

> `Warning: validateDOMNesting(...): <tr> cannot appear as a child of <div>.`

next/src/components/variables/form/VariableForm.tsx
chloe-renaud: > This allows text input without any indication that the field requires a number. The user can’t submit the variable, but may not understand why.

Ah yes, the type "number" does not work as well as I thought. Maybe we should simply use the [NumberField](https://base-ui.com/react/components/number-field) component to handle numbers.

> Also, the min can be greater than the max

I tried to handle it but I wasn't happy with the solution. There is already this issue in old Pogues so I'll keep it that way and I will improve it in a later update (with the NumberField component).

MailineN: In that case, it would be helpful to add a comment in the code and an error state on the field so the user can clearly see what’s wrong.
+Test

MailineN (changes requested):
For this kind of large PR, try to avoid addressing multiple topics at once. If it’s necessary, indicate it clearly in the PR description (for example: codeList, poguesModel, and scope here, even that the last two seems clearly related), along with what reviewers should focus on first

MailineN:
Another option would be to create two separate PRs and merge them together before deploying the feature for UAE, although I’m not really a fan of that

next/src/components/variables/form/VariableForm.tsx
chloe-renaud: I'd rather wrap the NumberField component than waste time on this since we'll delete it eventually tbf.

chloe-renaud:
> For this kind of large PR, try to avoid addressing multiple topics at once. If it’s necessary, indicate it clearly in the PR description (for example: codeList, poguesModel, and scope here, even that the last two seems clearly related), along with what reviewers should focus on first

But I didn't change anything in the functionality, I only harmonized in the app how I handled some Sonar issues in the Variables part (i.e. reduce the complexity of `? () : ()` by using `if`). It's standard programming, otherwise we'd have technical debt with a lot of components handling things differently in many parts of the app...

laurentC35 (approved):
On va s'arrêter là pour cette PR, ça devient de la sur qualité, je valide en l'état, merci !

Je vous propose de faire des TS ou directement d'autres PR pour mettre en place les corrections que vous suggérer

PR #1111, opened 2025-12-10
Various form improvements

chloe-renaud:
refactor: form improvements
- moved schema into their own files next to their respective ThingyForm components.
- created a Form component which handle the form dirty state, cancel/validate buttons and form style
- created a Field component which handle input labeling, error, description
- improved form ui components to have the same props, with more comments and test

feat: added input number
fix: use radio button in questionnaire form
style: form harmonization


QRuhier (approved):

next/src/components/variables/form/VariableForm.tsx
```diff
@@ -247,54 +277,74 @@ export default function VariableForm({
...
-                {...field}
-                onChange={(v) => {
-                  const value = v.target.value;
-                  field.onChange(value ? Number(value) : 0);
-                }}
-              />
+                name={name}
+                touched={isTouched}
+              >
+                <NumberField
```
QRuhier, line 342: Here it should add a param to enable only integer

next/src/components/variables/form/VariableForm.tsx
```diff
@@ -304,33 +354,28 @@ export default function VariableForm({
...
-              onChange={(v) => {
-                const value = v.target.value;
-                field.onChange(value ? Number(value) : 1);
-              }}
-            />
+              name={name}
+              required
+              touched={isTouched}
+            >
+              <NumberField
```
QRuhier, line 370: Same, it should enable only integer

QRuhier (approved):
Globally it seems good, but a lot of different changes so not sure if i missed something.

Just a few things about numbers concerning integers, but as we already did not handle it before in Pogues-Legacy, we can ignore it for now and handle it in another PR in the future

chloe-renaud:
> Globally it seems good, but a lot of different changes so not sure if i missed something.
> 
> Just a few things about numbers concerning integers, but as we already did not handle it before in Pogues-Legacy, we can ignore it for now and handle it in another PR in the future

Yeah deployment is on hold anyway so we'll have to time before 2.3.0 to see if there is any oversight.

PR #1115, opened 2025-12-17
fix: handle hyphen in all nomenclatures

QRuhier:
handle hyphen (`-`) in nomenclatures for suggesters, adding it into the regexp used for tokens


chloe-renaud (changes requested):
Changelog update to do.

chloe-renaud (approved):

PR #1120, opened 2026-01-08
feat(next): improve menu items UI

chloe-renaud:
Made menu label more accessible, fixed logout icon, display delete action in red and add "..." for actions which open a form as they need additional information to execute the action (e.g. "edit").

### DIFF
--- a/next/src/components/layout/User.tsx
+++ b/next/src/components/layout/User.tsx
@@ -13,7 +13,7 @@ interface UserProps {
 /** Compute initials of the current user to display as an avatar. */
 export default function User({ user }: Readonly<UserProps>) {
   const { t } = useTranslation();
-  const { isUserLoggedIn, logout } = useOidc();
+  const { isUserLoggedIn, logout = () => {} } = useOidc();
   const initials =
     user?.givenName && user?.familyName
       ? `${user.givenName.charAt(0).toUpperCase()}${user.familyName.charAt(0).toUpperCase()}`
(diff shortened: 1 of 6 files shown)


next/src/components/layout/User.tsx
```diff
@@ -27,10 +27,11 @@ export default function User({ user }: Readonly<UserProps>) {
 
   return isUserLoggedIn ? (
     <Menu
+      label="Open user navigation menu"
```
QRuhier, line 30: Actually the label will not be used considering the current implementation of `Menu` : here you have a children so the aria-label using this label is not rendered

next/src/components/ui/Menu.tsx
```diff
@@ -59,6 +69,7 @@ export default function Menu({ children, items, label }: Readonly<MenuProps>) {
                   </span>
                 )}
                 {item.label}
+                {item.type === MenuItemType.Form ? '...' : ''}
```
QRuhier: That seems a bit strange for me (especially when having many actions, it makes too much "..." in the menu. But i let the team give its opinion on it

QRuhier (dismissed):

next/src/components/layout/User.tsx
```diff
@@ -27,10 +27,11 @@ export default function User({ user }: Readonly<UserProps>) {
...
       items={[
         {
           label: t('common.logout'),
-          icon: <LogoutIcon />,
+          Icon: <LogoutIcon height="17px" width="17px" />,
```
QRuhier, line 34: Nice catch 😄

next/src/components/ui/Menu.tsx
MailineN: I agree with @QRuhier

MailineN (approved):

next/src/components/ui/Menu.tsx
chloe-renaud: I wanted to use the same logic as the browser to stay consistent.

Some actions will be direct and won't open a form (such as "delete" or "view"), so it will not have the "..." to indicate to the user they will have to specify stuff to complete their action.

PR #1121, opened 2026-01-08
feat(legacy): handle gender source variable for pairwise global variables

chloe-renaud:


src/constants/dictionary/pairing.ts
```diff
@@ -7,19 +7,19 @@ export const pairingDictionary: Dictionary = {
     en: 'Household members pairing',
   },
   pairingSourceVariable: {
-    fr: 'Variable source',
-    en: 'Source variable',
+    fr: 'Variable source du nom',
+    en: 'Name source variable',
```
QRuhier: It's more first name (as the business analysts confirmed)

QRuhier (changes requested):
The transformation stateToRemote is missing here (you still only persist Scope, not `pairingSourceVariable` object).

If i'm right, it's directly here : https://github.com/InseeFr/Pogues/blob/next/src/model/transformations/component.jsx#L730

Else it's nice !

src/constants/dictionary/pairing.ts
chloe-renaud: I'd rather keep the label as short as possible. Let's wait and see if business analysts deem necessary to change it or if it seems obvious to them for now.

QRuhier (approved):

src/constants/dictionary/pairing.ts
JulienCarmona: Yes we prefer "Variable source du Prénom", cause it's the main (and maybe only but can't be sure) use

laurentC35: @JulienCarmona But who are you ? 😜

src/model/transformations/response-format-pairing.jsx
```diff
@@ -10,13 +10,20 @@ export function remoteToState(remote) {
         id,
       },
     ],
-    scope,
+    sourceVariableReferences,
+    scope, // deprecated
```
laurentC35, line 14: Great ! 🚀 
We have to keep it because of generation of pogues-model -> DDI in [EnoXML](https://github.com/InseeFr/Eno-Xml)

ToDo: handle new modeling for this transformation (pogues->DDI) in https://github.com/InseeFr/Eno-Xml

laurentC35 (approved):

PR #1126, opened 2026-01-21
chore(build): pnpm migration

MailineN:


next/src/components/layout/Header.test.tsx
```diff
@@ -60,7 +60,7 @@ describe('Header', () => {
       expect.objectContaining({
         user: user,
       }),
-      expect.anything(),
+      undefined,
```
QRuhier, line 65: Can you put here the doc you found explaining this change ?

QRuhier (approved):

next/src/components/layout/Header.test.tsx
QRuhier: I found this : https://dev.to/peterlidee/how-to-test-if-a-mocked-component-receives-the-correct-props-in-jest-and-rtl-new-since-react-19-2240

So we remember we now always need to set `undefined` as 2nd param

laurentC35 (changes requested):
Since react and react-dom are shared with pogues-legacy (federation module)

We must have the same react & react-dom version in both app (pogues & pogues-legacy)

Sorry 😢

laurentC35:
Deprecated, already did in https://github.com/InseeFr/Pogues/pull/1136

PR #1131, opened 2026-02-18
feat: enable only text type for external variables

QRuhier:
- When creating an external variable, datatype is now "text" and cannot be changed
- We do not change the existing external variables. So when editing an external variable, if its type is not "text", the user can keep the current type or change it to "text"

### DIFF
--- a/next/CHANGELOG.md
+++ b/next/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- An external variable can now only have a "text" type.
+
 ## [2.3.0](https://github.com/InseeFr/Pogues/releases/tag/2.3.0) - 2026-01-19
 
 ### Added
--- a/next/CHANGELOG.md
+++ b/next/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- An external variable can now only have a "text" type.
+- An external variable can now only have a "text" datatype.
 
 ## [2.3.0](https://github.com/InseeFr/Pogues/releases/tag/2.3.0) - 2026-01-19
 
--- a/next/src/components/ui/form/Select.tsx
+++ b/next/src/components/ui/form/Select.tsx
@@ -60,7 +60,7 @@ export default function Select<T>({
           select-none
           focus-visible:outline-2 focus-visible:-outline-offset-1 focus-visible:outline-primary
           active:bg-accent data-popup-open:bg-accent
-          data-disabled:cursor-not-allowed data-disabled:opacity-50
+          data-disabled:cursor-not-allowed data-disabled:opacity-50 data-disabled:hover:bg-default
         `}
       >
         <BaseUISelect.Value />
--- a/next/src/components/ui/form/Select.tsx
+++ b/next/src/components/ui/form/Select.tsx
@@ -52,14 +52,15 @@ export default function Select<T>({
         className={`
           flex min-w-36 w-full
           text-sm text-default font-normal
-          bg-default
+          cursor-pointer
+          bg-default hover:bg-main
           items-center justify-between gap-3
           rounded-md border border-default
           pr-3 pl-3.5 py-4
           select-none
           focus-visible:outline-2 focus-visible:-outline-offset-1 focus-visible:outline-primary
           active:bg-accent data-popup-open:bg-accent
-          ${disabled ? 'cursor-not-allowed opacity-50' : 'cursor-pointer hover:bg-main'}
+          data-disabled:cursor-not-allowed data-disabled:opacity-50
         `}
       >
         <BaseUISelect.Value />
--- a/next/src/components/ui/form/Select.tsx
+++ b/next/src/components/ui/form/Select.tsx
@@ -14,6 +14,8 @@ type Props<T> = {
   defaultValue?: BaseUISelect.Root.Props<T>['defaultValue'];
   /** Currently selected value (if controlled). */
   value?: BaseUISelect.Root.Props<T>['value'];
+  /** Disable option selection. */
+  disabled?: boolean;
 };
 
 /**
@@ -36,6 +38,7 @@ export default function Select<T>({
   options = [],
   defaultValue,
   value,
+  disabled = false,
 }: Readonly<Props<T>>) {
   return (
     <BaseUISelect.Root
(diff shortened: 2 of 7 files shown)


next/src/components/ui/form/Select.tsx
```diff
@@ -36,26 +38,29 @@ export default function Select<T>({
...
-          bg-default hover:bg-main
+          bg-default
           items-center justify-between gap-3
           rounded-md border border-default
           pr-3 pl-3.5 py-4
           select-none
           focus-visible:outline-2 focus-visible:-outline-offset-1 focus-visible:outline-primary
           active:bg-accent data-popup-open:bg-accent
-        "
+          ${disabled ? 'cursor-not-allowed opacity-50' : 'cursor-pointer hover:bg-main'}
```
chloe-renaud: We can use `data-disabled:something` instead

See https://base-ui.com/react/components/select

next/src/components/variables/form/VariableForm.test.tsx
```diff
@@ -123,4 +125,110 @@ describe('VariableForm', () => {
...
+    );
+
+    const datatypeSelect = getByRole('combobox', { name: /datatype/i });
+
+    await waitFor(() => {
+      expect(datatypeSelect).toBeDisabled();
+    });
+  });
+
+  it('should disable datatype.typeName select when editing an external variable with text datatype typename', async () => {
```
chloe-renaud: C'est la même chose que le test au-dessus -> à supprimer.

next/src/components/variables/form/VariableForm.test.tsx
```diff
@@ -123,4 +125,110 @@ describe('VariableForm', () => {
     expect(await findAllByRole('alert')).toHaveLength(1);
     expect(getByText('You must provide a description')).toBeDefined();
   });
+
+  it('should disable datatype.typeName select by default when variable is external', async () => {
```
chloe-renaud:
```suggestion
  it('should disable datatype selection when variable is external (it can only be text)', async () => {
```

next/src/components/variables/form/VariableForm.test.tsx
```diff
@@ -123,4 +125,110 @@ describe('VariableForm', () => {
...
+    const { getByRole } = await renderWithRouter(
+      <VariableForm
+        questionnaireId="q-id"
+        onSubmit={vi.fn()}
+        submitLabel="Validate"
+        scopes={new Map<string, string>()}
+      />,
+    );
+
+    const datatypeSelect = getByRole('combobox', { name: /datatype/i });
```
chloe-renaud:
```suggestion
    expect(getByRole('combobox', { name: /datatype/i })).toBeDisabled();
```

I believe we can do the assertion in one line instead.

And maybe we could add a check "it is set on 'text'".

next/src/components/variables/form/VariableForm.test.tsx
```diff
@@ -123,4 +125,110 @@ describe('VariableForm', () => {
...
+
+    await waitFor(() => {
+      expect(datatypeSelect).toBeDisabled();
+    });
+  });
+
+  it('should allow as datatype options text/date/numeric/boolean for a calculated variable', async () => {
```
chloe-renaud:
```suggestion
  it('should allow all datatype options for calculated variables', async () => {
```

next/src/components/variables/form/VariableForm.test.tsx
```diff
@@ -123,4 +125,110 @@ describe('VariableForm', () => {
...
+    const datatypeSelect = getByRole('combobox', { name: /datatype/i });
+    expect(datatypeSelect).toBeEnabled();
+
+    // open select
+    fireEvent.click(datatypeSelect);
+    const options = await findAllByRole('option');
+
+    // allowed options : text, date, numeric, boolean
+    expect(options).toHaveLength(4);
+    expect(options[0]).toHaveTextContent(/text/i);
```
chloe-renaud: I'm not a fan of manually telling "options[0]" is text (what if we change the order of the options? it shouldn't change the test, the user doesn't care and that is not what we want to test).

Can't we do something like this instead: (not sure about the syntax)

```js
expect(screen.getByRole("option", { name: "/date/i"})).toBeInTheDocument();
```

next/src/components/variables/form/VariableForm.test.tsx
```diff
@@ -123,4 +125,110 @@ describe('VariableForm', () => {
...
+    const datatypeSelect = getByRole('combobox', { name: /datatype/i });
+    expect(datatypeSelect).toBeEnabled();
+
+    // open select
+    fireEvent.click(datatypeSelect);
+    const options = await findAllByRole('option');
+
+    // allowed options : text and current variable datatype typeName
+    expect(options).toHaveLength(2);
+    expect(options[0]).toHaveTextContent(/text/i);
```
chloe-renaud: Ditto.

next/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- An external variable can now only have a "text" type.
```
chloe-renaud:
```suggestion
- An external variable can now only have a "text" datatype.
```

chloe-renaud (changes requested):
Code seems good, but I have a few remarks about the tests.

next/src/components/variables/form/VariableForm.test.tsx
QRuhier: ups, just forgot here i wanted to add clearly the variable prop with text datatype (editing an existing variable).
I let you tell me if it's ok now or if you think it's not usefull

MailineN (approved):

next/src/components/ui/form/Select.tsx
```diff
@@ -52,14 +52,15 @@ export default function Select<T>({
...
-          bg-default
+          cursor-pointer
+          bg-default hover:bg-main
           items-center justify-between gap-3
           rounded-md border border-default
           pr-3 pl-3.5 py-4
           select-none
           focus-visible:outline-2 focus-visible:-outline-offset-1 focus-visible:outline-primary
           active:bg-accent data-popup-open:bg-accent
-          ${disabled ? 'cursor-not-allowed opacity-50' : 'cursor-pointer hover:bg-main'}
```
chloe-renaud: ptete data-disabled:hover:bg-default ?

chloe-renaud (commented):
Parfait pour moi les tests, ptete juste un oubli sur un style

next/src/components/ui/form/Select.tsx
QRuhier: Petit loupé de copier-coller :D

chloe-renaud (approved):

PR #1117, opened 2025-12-23
refactor: VTL editor form improvements

chloe-renaud:

PR #1132, opened 2026-02-24
fix: codesList form crashes when form errors

QRuhier:
missing prop for setting VTL errors in codesList form, since #1117 , making form crash when there are errors


chloe-renaud:
LGTM

PR #1137, opened 2026-03-20
feat: description filter label

MailineN:


src/constants/dictionary.ts
```diff
@@ -322,8 +322,8 @@ const dictionary: Dictionary = {
     fr: 'Sélectionnez une unité de mesure',
   },
   description: {
```
QRuhier: Can you rename the key ? So that we understand it's precisely used for filters description, not a common key we could reuse

QRuhier (changes requested):

laurentC35 (approved):

PR #1140, opened 2026-03-26
fix: ts build

MailineN:


next/src/api/personalization.test.ts
```diff
@@ -66,13 +66,14 @@ it('Get personalization data works', async () => {
   expect(res).toEqual(mockData);
 });
 
-it('Delete personalization data works', async () => {
+it('Delete personalization data throws on 404', async () => {
```
chloe-renaud: Why is the 204 test replaced?

next/src/components/variables/form/VariableForm.tsx
```diff
@@ -27,7 +28,7 @@ type Props = {
   /** Label to display on the submit button */
   submitLabel: string;
   /** Available scopes with the mapping between id and name. */
-  scopes: Map<string, string>;
+  scopes?: Scopes;
```
chloe-renaud: I'd rather remove `Scopes` type entirely and replace it with `Map<string, string>` everywhere.

next/src/components/questionnaires/create/CreateQuestionnaireForm.tsx
```diff
@@ -1,13 +1,14 @@
-import { useMutation, useQueryClient } from '@tanstack/react-query';
-import { useNavigate } from '@tanstack/react-router';
-import toast from 'react-hot-toast';
-import { useTranslation } from 'react-i18next';
+import { useMutation, useQueryClient } from "@tanstack/react-query";
```
chloe-renaud: The `''` should not be replaced with `""`. Check the prettier configuration.

chloe-renaud (changes requested):

MailineN:
Feel free to discuss the prettier conf, I also added a lib to automatically sort import (I can remove it if you don't want it)

QRuhier (approved):

PR #1145, opened 2026-04-14
fix(legacy): crash when codeList not in store

QRuhier:
Crash when loading a questionnaire, with a question using a codeList that is not in the codesLists store in the model.
(bug added by UCQ based on a variable)

Happens currently when we have simultaneously multiple versions of a same nomenclature (since nomenclature are not well handled for now, the old versions are sometimes removed from the store in the model...)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Questionnaire can now be loaded if a question uses a code list (or nomenclature) that does not exist anymore. It can happen if an existing question used a previous version of a nomenclature.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Questionnaire can now be loaded if a question uses a code list (or nomenclature) that does not exist anymore. It can happen if an existing question used a previous version of a nomenclature.
```
chloe-renaud:
```suggestion
- Questionnaire loading no longer crash when a question uses a code list (or nomenclature) that does not exist anymore. It could happen when an existing question used a previous version of a nomenclature.
```

chloe-renaud (changes requested):

chloe-renaud (approved):

CHANGELOG.md
nsenave: typo: crashes

PR #1151, opened 2026-04-20
feat: add filter for ucq based on a variable

QRuhier:


QRuhier:
Needs to be done in Pogues-bo & Eno before merging
Or we can add a feature flag for not showing the optionFilter field in form (no risk of regression)

laurentC35:
⚠️ pogues-model in version 1.15.5 is required in pogues-api and Eno !

need somes adpatation in Eno

QRuhier:
@laurentC35 Yeah i did https://github.com/InseeFr/Eno/pull/1384
But when checking quickly with @nsenave it seems we have issue with ddi so it would need some changes

QRuhier:
In the end we did not change the model : for ddi reason, we keep `OptionFilter` in the Question, assuming we handle it only for UCQ.
Then we disabled the field for filtering in table.

chloe-renaud (approved):
Nice, love the comments, tests and changelog :)

PR #1161, opened 2026-05-22
feat: add contact mail env var

MailineN:


laurentC35 (changes requested):
The environment variable is missing from the `.env` file.

Everything else is perfect

laurentC35 (approved):

PR #1167, opened 2026-06-05
feat: use nomenclatures end point

laurentC35:
Need the end-point from the backend (api/nomenclatures)

Feel free to improve it


QRuhier (approved):
Wonderful !

PR #1169, opened 2026-06-11
fix: variables version scopes

laurentC35:


QRuhier (approved):

chloe-renaud:
What does it fix exactly? Was the page crashing before when, no scopes were provided?

PR #1175, opened 2026-07-06
feat: handle validation errors from API

laurentC35:


QRuhier (approved):
That was fun to do this :-D

PR #1180, opened 2026-07-17
fix(legacy): remove depreacted dataCollection reference

laurentC35:


MailineN (changes requested):
You forgot to update the changelog, but other than that, it looks good to me

MailineN (approved):

PR #1186, opened 2026-08-04
refactor: series are now optionals in details form

MailineN:


package.json
```diff
@@ -20,6 +20,7 @@
   "devDependencies": {
     "@eslint/js": "^9.17.0",
     "@trivago/prettier-plugin-sort-imports": "^6.0.2",
+    "lodash-es": "^4.17.21",
```
laurentC35: Do you really need this dependency ?

MailineN: It wasn't building without it (I probably messed up something trying to bump packages version sorry ˆˆ)

laurentC35: Ok, we have to remove it later

laurentC35:
Apart from the issue of dependencies, it’s ok for me

PR #1196, opened 2026-08-06
feat(fix): short name field in questionnaire details page

MailineN:


MailineN:
Don't mind the duplication, it is because I just added a field to the form

laurentC35 (changes requested):
You have to control name value.

Like in this form https://github.com/InseeFr/Pogues/blob/main/next%2Fsrc%2Fcomponents%2Fvariables%2Fform%2FVariableForm.tsx#L190

MailineN:
> You have to control name value.
> 
> Like in this form https://github.com/InseeFr/Pogues/blob/main/next%2Fsrc%2Fcomponents%2Fvariables%2Fform%2FVariableForm.tsx#L190

I added a form validation check instead of a conversion. I think it's less surprising for the user if they notice that the saved value matches their input.

PR #1238, opened 2026-08-22
feat: update VTL editor, align legacy on React 19, add standalone mode

NicoLaval:
- Standalone mode to run without Keycloak/API (in-memory mock, `dev:standalone` / `pnpm standalone`, docs updated)
- Bump `@making-sense/antlr-editor` to 2.9.4 to fix VTL script validation errors in the UI (empty / partial expressions)
- Legacy aligned on React 19 with next (Module Federation, Monaco / VTL 2.1)
- Cleanup: drop enzyme and obsolete specs; small CSS fixes (fonts / Tailwind layers)
- Next: mock API wiring, VTL editor, and legacy remote follow the same stack
- Skip SonarCloud/SonarQube CI jobs on fork PRs (no Sonar token); still run on InseeFr pushes and internal PRs


laurentC35:
ça a l'air en good en première lecture !
Le passage à React 19 nous enlèverai une épine du pied !

Cependant, quelques fonctionnalité ne fonctionnent plus:
- "Voir le detail" d'une question fait cracher l'application (legacy seul ou next & legacy ensemble)
- De manière générale, le formulaire pour ajouter une Question ne fonctionne pas
- En mode next+legacy ensemble, la barre permettant l'ajout de Question, Séquence, Boucle & Filtre, Sauvegarde, etc disparait

NicoLaval:
@laurentC35, thanks for the review.

Root cause: under React 19, `defaultProps` on function components (and on `forwardRef` / `connect` / `reduxForm` HOCs) are no longer applied.

Fixes:
- migrate `defaultProps` to defaults in destructuring
- add a `reduxFormNamed` wrapper that passes `form` explicitly
- restore defaults / explicit props on containers that still relied on `defaultProps`

Keep me informed.

NicoLaval:
Hello,

A few extra commits on top of the VTL / React 19 work:

- bumped dependencies
- added a way to run against a local Pogues-API without Keycloak (pnpm dev:api / preview:api), with README notes
- documented that dropping a questionnaire file with the same id overwrites the one in the DB

I also merged main (3.8.0) to clear the conflicts. We kept our stack and took the releases page from main.

laurentC35:
@NicoLaval 
Bad news, for generation purpose and historical reason.
It is requested that the dollar symbol '$' be permitted. (to detect variable for DDI géneration), based on XSLT.

That said, this change is desirable in the long term, and the team is excited at the prospect of doing away with the dollar signs around variables.

However, this will be done at a later stage. Not during the migration to React 19.

We need to examine the entire build chain (DDI, Lunatic with Eno) to properly assess the impact of removing the dollar signs.

In the meantime, is it possible to have some sort of exception for the dollar sign?

This could be done either by overriding the VTL editor in Pogues directly (without breaking the new version of the editor)?

Or, taking an even more ‘brutal’ approach, we could implement a way to detect the specific dollar sign error and ignore it (make it non-blocking) when validating a form using the VTL editor.

NicoLaval:
Agreed, we will not remove `$VAR$` as part of the React 19 / VTL editor migration.

I added a commit to temporary have a silent Pogues-side exception (legacy + next) so now, `$` VTL lexer/parser errors are non-blocking for form validation. Stored values keep `$VAR$` as today, we did not fork/override the VTL grammar in the editor package.

I also added [an issue](https://github.com/InseeFr/Pogues/issues/1246) to help you to give up this trick.

PR #1245, opened 2026-09-17
feat: prevent release request when there is already one for the current version

MailineN:


laurentC35 (changes requested):
Mini review

For getting id of latest version, you should use this end-point : _`/api/persistence/questionnaire/{poguesId}/version/last`

It returns something like that

```json
{
  "id": "0faeeafa-d715-43c7-beeb-d83d421eb2dd",
  "poguesId": "poguesId",
  "timestamp": "2026-09-17T08:48:41.235219Z",
  "day": "2026-09-17T00:00:00.000Z",
  "data": null,
  "author": "guest"
}
```

It's more efficient than getting all versions

QRuhier (commented):
Except for getting last version from existing endpoint, good for me !

QRuhier (approved):

